PR #39, opened 2019-06-03
Move grammar datatypes to the language packages

robrix:
This PR moves the definitions of the grammar datatypes into the language packages.

Depends on https://github.com/tree-sitter/haskell-tree-sitter/pull/112


robrix:
https://en.wikipedia.org/wiki/Joyeuse

PR #48, opened 2019-06-04
Fix "programing" typo

hectorj:


hectorj:
[It seems both "programming" and "programing" are acceptable](https://english.stackexchange.com/questions/244280/programming-versus-programing-which-is-preferred), but this is the only occurence of the single "m" variant in this document while the double "m" appears 6 times

PR #99, opened 2019-06-06
using Attoparsec for git ls-tree result

yuanw:
## GOAL
address https://github.com/github/semantic/issues/34

not quite sure to add tests around new functions, since they are not exported

### DIFF
--- a/src/Semantic/Git.hs
+++ b/src/Semantic/Git.hs
@@ -11,10 +11,12 @@ module Semantic.Git
   , OID(..)
   ) where
 
-import Control.Monad.IO.Class
-import Data.Text as Text
-import Shelly hiding (FilePath)
-import System.IO (hSetBinaryMode)
+import           Control.Monad.IO.Class
+import           Data.Char (isSpace)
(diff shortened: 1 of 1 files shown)


src/Semantic/Git.hs
```diff
@@ -30,12 +32,25 @@ catFile gitDir (OID oid) = sh $ do
 lsTree :: FilePath -> OID -> IO [TreeEntry]
 lsTree gitDir (OID sha) = sh $ do
   out <- run "git" [pack ("--git-dir=" <> gitDir), "ls-tree", "-rz", sha]
-  pure $ mkEntry <$> splitOn "\NUL" out
-  where
-    mkEntry row | [mode, ty, rest] <- splitOn " " row
-                , [oid, path] <- splitOn "\t" rest
-                = TreeEntry (objectMode mode) (objectType ty) (OID oid) (unpack path)
-                | otherwise = nullTreeEntry
+  pure $ either mempty id (Parser.parseOnly treeEntriesParser out)
```
yuanw: not quite sure how to handle failure cases

PR #102, opened 2019-06-06
Split TypeScript assignment into TSX and TypeScript

joshvera:
In https://github.com/tree-sitter/tree-sitter-typescript/pull/68, we separated the TSX and TypeScript syntax in order to support type arguments in TSX and avoid ambiguity.

This PR adds a new assignment to semantic and moves `JSX`, `TSX`, and `JavaScript` to this new parser/assignment in order to assign more precisely. In practice this means type assertions are no longer supported in Flow/JS and JSX fragments are no longer parsed in TypeScript.


src/Language/TSX/Assignment.hs
```diff
@@ -0,0 +1,971 @@
...
+  where
+    everything = choice [
+      asExpression,
+      nonNullExpression',
+      importAlias',
+      internalModule,
+      super,
+      object,
+      array,
+      jsxElement',
```
joshvera: I've had to duplicate typescript assignment in order to leave `jsxElement'` and `jsxFragment` in here. I could move assignment out into some `Common` module and pass in a list of terms to mappend instead. Thoughts?

rewinfrey: Thanks for calling out the duplication. Given the size of this patch as it is, I'd be in favor of merging this and to extract a common assignment module as a follow up PR. That might also make a weightier issue for someone in the community to take on. Either way I'm 👍

src/Language/TSX/Resolution.hs
```diff
@@ -0,0 +1,128 @@
...
+                  <> maybe mempty (:[]) packageDotJSON
+                  <> (((path </> "index") <.>) <$> exts)
+  trace ("searching in " <> show searchPaths)
+  maybe (Left searchPaths) Right <$> resolve searchPaths
+
+typescriptExtensions :: [String]
+typescriptExtensions = ["ts", "tsx", "d.ts"]
+
+javascriptExtensions :: [String]
+javascriptExtensions = ["js"]
```
rewinfrey: Are there differences between this `Resolution` module and `TypeScript.Resolution`? I'm curious if we can still maintain a single `Resolution` module for `TypeScript` and `TSX`?

test/Examples.hs
```diff
@@ -83,7 +83,8 @@ languages :: [LanguageExample]
 languages =
   [ le "python" ".py" "examples" (Just "script/known_failures.txt")
   , le "ruby" ".rb" "examples" (Just "script/known_failures.txt")
-  , le "typescript" ".ts" "examples" (Just "script/known_failures.txt")
+  , le "typescript" ".ts" "examples" (Just "typescript/script/known_failures.txt")
+  , le "typescript" ".tsx" "examples" (Just "typescript/script/known_failures.txt")
```
rewinfrey: I'm not familiar with this context, but is `"typescript"` as the language correct here rather than `"tsx"`?

joshvera: Yeah `"typescript"` refers to the directory the example repos are in. This adds about 150 tests to `parse-examples`.

PR #149, opened 2019-06-14
Enable warnings in tests

jhrcek:
Still work in progress on https://github.com/github/semantic/issues/127.
Checking how this will behave in CI.


jhrcek:
I'm going to fix some of those warnings in tests in the followup PR

PR #184, opened 2019-06-26
Scope safety

robrix:
This PR is an experiment in using a scope-safe definition of `Core` which should nevertheless retain the connection with surface language symbols. This will be accomplished by simply retaining the user-provided name (if any) in the `Core` itself, and pushing these into an environment of sorts.

I’m also hoping to once more split the treatment of global vs. local symbols s.t. we can read the environment/context directly as a stack of assumptions.

- [x] Depends on #195.


semantic-core/src/Data/Scope.hs
```diff
@@ -0,0 +1,102 @@
...
+match f x = either Z (S . pure) (f x)
+
+matchMaybe :: (b -> Maybe a) -> (b -> Either a b)
+matchMaybe f a = maybe (Right a) Left (f a)
+
+incr :: (a -> c) -> (b -> c) -> Incr a b -> c
+incr z s = \case { Z a -> z a ; S b -> s b }
+
+
+newtype Scope a f b = Scope { unScope :: f (Incr a (f b)) }
```
robrix: `Scope` embeds an `f` with a bound variable, represented as `f` wrapping `Incr` wrapping `f`.

1. The outer `f` is to embed subterms, necessary for any reasonable binding syntax.

2. `Incr` is used to ensure that the type of the embedded term differs from the type of the embedding term; “nested datatypes” of this nature enable us to use the type system to ensure that we aren’t unwrapping scopes without correctly handling their bound variables, which is key for ensuring the correctness of finicky operations like capture-avoiding substitution, calculating free variables, and so on.

3. The inner `f` enables us to instantiate a `Scope`’s bound variable with an entire argument _without_ changing that argument; in effect, instead of pushing the `Incr`’s `S` constructors all the way through the argument to wrap each new bound variable, we wrap entire subterms with `S` instead. This is more efficient, tho it also makes certain folds kind of painful.

Using this inside syntactic binders like `Lam` is what gives us scope safety!

semantic-core/src/Data/Scope.hs
```diff
@@ -0,0 +1,102 @@
...
+  deriving (Foldable, Functor, Traversable)
+
+instance HFunctor (Scope a) where
+  hmap f = Scope . f . fmap (fmap f) . unScope
+
+instance (Eq   a, Eq   b, forall a . Eq   a => Eq   (f a), Monad f) => Eq   (Scope a f b) where
+  (==) = (==) `on` fromScope
```
robrix: Unlike “normal” de Bruijn-indexed representations (`f (Incr a)`), Paterson’s extended representation (`f (Incr (f a))`) can represent alpha-equivalent terms in multiple ways. We therefore define specialized `Eq` and `Ord` instances so that `==` and `compare` respect alpha-equivalence.

`fromScope` can be seen as canonicalizing the extended representation to the basic one.

semantic-core/src/Data/Scope.hs
```diff
@@ -0,0 +1,102 @@
...
+
+fromScope :: Monad f => Scope a f b -> f (Incr a b)
+fromScope = unScope >=> sequenceA
+
+toScope :: Applicative f => f (Incr a b) -> Scope a f b
+toScope = Scope . fmap (fmap pure)
+
+
+-- | Bind occurrences of a variable in a term, producing a term in which the variable is bound.
+bind1 :: (Applicative f, Eq a) => a -> f a -> Scope () f a
```
robrix: `bind1`, `bind`, and `bindEither` walk a term, replacing some/all free variables with bound ones in a new `Scope`.

semantic-core/src/Data/Scope.hs
```diff
@@ -0,0 +1,102 @@
...
+
+bind :: Applicative f => (b -> Maybe a) -> f b -> Scope a f b
+bind f = bindEither (matchMaybe f)
+
+bindEither :: Applicative f => (b -> Either a c) -> f b -> Scope a f c
+bindEither f = Scope . fmap (match f) -- FIXME: succ as little of the expression as possible, cf https://twitter.com/ollfredo/status/1145776391826358273
+
+
+-- | Substitute a term for the free variable in a given term, producing a closed term.
+instantiate1 :: Monad f => f b -> Scope a f b -> f b
```
robrix: `instantiate1`, `instantiate`, and `instantiateEither` walk through a `Scope`, substituting some/all variables with new terms and eliminating the `Scope`.

semantic-core/src/Data/Scope.hs
```diff
@@ -0,0 +1,102 @@
...
+instantiate1 t = instantiate (const t)
+
+instantiate :: Monad f => (a -> f b) -> Scope a f b -> f b
+instantiate f = instantiateEither (either f pure)
+
+instantiateEither :: Monad f => (Either a b -> f c) -> Scope a f b -> f c
```
robrix: Note that `instantiateEither` is able to change the type of the free variables. This can be used to prove that there _are_ no free variables (e.g. by instantiating `c` to `Void`; tho note that this is more easily done with `traverse`). I’m personally hoping to use it to replace bound variables with information from the evaluating environment, and therefore to potentially avoid lookups in an environment.

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+import Control.Effect.Carrier
+import Control.Monad (ap)
+import Control.Monad.Module
+import Data.Coerce (coerce)
+import Data.Functor.Const
+import Data.Scope
+
+data Term sig a
+  = Var a
+  | Term (sig (Term sig) a)
```
robrix: The higher-order free monad…

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+deriving instance ( Eq a
+                  , RightModule sig
+                  , forall g x . (Eq  x, Monad g, forall y . Eq  y => Eq  (g y)) => Eq  (sig g x)
+                  )
+               => Eq  (Term sig a)
+deriving instance ( Ord a
+                  , RightModule sig
+                  , forall g x . (Eq  x, Monad g, forall y . Eq  y => Eq  (g y)) => Eq  (sig g x)
+                  , forall g x . (Ord x, Monad g, forall y . Eq  y => Eq  (g y)
+                                                , forall y . Ord y => Ord (g y)) => Ord (sig g x)
```
robrix: …now with horrifying higher-order quantified constraints!

1. I didn’t want to use `Eq1`, `Ord1`, `Show1`, etc., because they’re awful, and don’t provide derivations.

2. `Core` (and any other syntax which binds variables) applies its type parameter `f` to more than one different type: `f a` for normal subterm positions` and `Scope () f a` (isomorphic to `f (Incr () (f a))`) for binders like `Lam`. Therefore, `Core` uses quantified constraints for its `Eq` instance to state that it can be equated whenever `f x` can be equated for all `x` that can be equated.

3. `Term` only ever applies its type parameter `sig` to `Term sig` and `a`; however, presumably due to `Core`’s quantified constraints, using the simpler constraint of e.g. `(Show a, forall x . Show x => Show (sig (Term sig) x))` causes the constraint solver to infinite loop. `UndecidableInstances` means business in this case. Therefore, we have to use the dramatically more complicated constraint `(Show a, forall g x . (Show x, forall y . Show y => Show (g y)) => Show (sig g x))`, which quantifies over some arbitrary type parameter `g`.

4. Both the `Eq` & `Ord` instances require `Monad` instances for `g` because `Core` requires a `Monad` instance for its parameter to be equated (due to the use of `Scope`). Likewise, they both require a `RightModule` instance for `sig`, because `Term` requires a `RightModule` instance for `sig` in its `Monad` instance. So these instances compile fine if you :fire: the `RightModule` & `Monad` constraints, but then they can’t actually be instantiated to work with `Core`.

5. The `Ord` instance duplicates the quantified `Eq` constraint (just like `Core`’s does) because of contravariance: `forall x . Ord x => c` is strictly smaller than `forall x . Eq x => c`, because there are fewer `Ord` instances than `Eq` instances. Therefore, it’s not possible to prove `forall x . Eq x => Eq (f x)` from `forall x . Ord x => Ord (f x)`, because the domain of the smaller is too small to prove its premise.

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+                  )
+               => Ord (Term sig a)
+deriving instance (Show a, forall g x . (Show x, forall y . Show y => Show (g y)) => Show (sig g x)) => Show (Term sig a)
+
+deriving instance ( forall g . Foldable    g => Foldable    (sig g)) => Foldable    (Term sig)
+deriving instance ( forall g . Functor     g => Functor     (sig g)) => Functor     (Term sig)
+deriving instance ( forall g . Foldable    g => Foldable    (sig g)
+                  , forall g . Functor     g => Functor     (sig g)
+                  , forall g . Traversable g => Traversable (sig g)) => Traversable (Term sig)
```
robrix: We don’t get an infinite loop from using more specialized constraints for these instances:

```haskell
deriving instance Foldable    (sig (Term sig)) => Foldable    (Term sig)
deriving instance Functor     (sig (Term sig)) => Functor     (Term sig)
deriving instance Traversable (sig (Term sig)) => Traversable (Term sig)
```

However, these are exactly the sort of undecidable instances that quantified constraints were designed to avoid, and given that we’ve already courted undecidability above, I felt it best to play it safe.

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+
+instance RightModule sig => Monad (Term sig) where
+  Var  a >>= f = f a
+  Term t >>= f = Term (t >>=* f)
+
+instance RightModule sig => Carrier sig (Term sig) where
+  eff = Term
+
+
+class (HFunctor sig, forall g . Functor g => Functor (sig g)) => Syntax sig where
```
robrix: This class is necessary if we want to fold over syntax, and unfortunately we currently still do kind of want to fold over syntax. I’m hoping to alleviate that, however.

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+             -> (a -> m b)
+             -> sig f a
+             -> sig n b
+
+instance Syntax (Scope ()) where
+  foldSyntax go bound free = Scope . go (bound . fmap (go free)) . unScope
+
+instance (Syntax l, Syntax r) => Syntax (l :+: r) where
+  foldSyntax go bound free (L l) = L (foldSyntax go bound free l)
+  foldSyntax go bound free (R r) = R (foldSyntax go bound free r)
```
robrix: This simple instance, distributing `foldSyntax` through both sides of a `:+:`, also hints at the motivation for using `Term` in the first place:

1. I intend to abstract it out into a `fused-effects-syntax` (or similar) package, à la `bound` (but with `QuantifiedConstraints` and more machinery for working with `fused-effects` specifically).

2. It allows us to decompose syntax like `Core`. A future PR will likely split the primitives out into their own type altogether, and thus evaluation will occur over e.g. `Term (Core :+: Prim) Name`.

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+
+
+iter :: forall m n sig a b
```
robrix: Bird & Paterson; Hinze; Martin, Bayley & Gibbons; and I all want you to know that this function was a _tremendous_ pain in the ass to define.

I’m quite proud that I was able to define this, at least, without either `Coercible` or quantified constraints.

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+instance RightModule sig => Carrier sig (Term sig) where
+  eff = Term
+
+
+class (HFunctor sig, forall g . Functor g => Functor (sig g)) => Syntax sig where
+  foldSyntax :: (forall x y . (x -> m y) -> f x -> n y)
+             -> (forall a . Incr () (n a) -> m (Incr () (n a)))
+             -> (a -> m b)
+             -> sig f a
+             -> sig n b
```
robrix: It’s really more of a `map`? But it takes operands like a `fold`? I don’t know.

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+iter :: forall m n sig a b
+     .  Syntax sig
+     => (forall a . m a -> n a)
+     -> (forall a . sig n a -> n a)
+     -> (forall a . Incr () (n a) -> m (Incr () (n a)))
+     -> (a -> m b)
+     -> Term sig a
+     -> n b
+iter var alg bound = go
+  where go :: forall x y . (x -> m y) -> Term sig x -> n y
```
robrix: It’s Mendler-style recursion! Or iteration! Or whatever!

So I guess Mendler also wants you to know that this function was a tremendous pain in the ass to define.

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+        go free = \case
+          Var a -> var (free a)
+          Term t -> alg (foldSyntax go bound free t)
+
+cata :: Syntax sig
```
robrix: This is a simplified take on `iter` for when you don’t care about the variable type parameter (nested or otherwise) and just want to produce a non-indexed value, e.g. pretty-printing.

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+     -> (forall a . sig (Const b) a -> b)
+     -> (Incr () b -> a)
+     -> (x -> a)
+     -> Term sig x
+     -> b
+cata var alg k h = getConst . iter (coerce var) (Const . alg) (coerce k) (Const . h)
+
+
+interpret :: (Carrier sig m, Member eff sig, Syntax eff) => (forall a . Incr () (m a) -> m (Incr () (m a))) -> (a -> m b) -> Term eff a -> m b
+interpret = iter id send
```
robrix: `Term` really _is_ the higher-order free monad, and as such we can interpret its contents as higher-order effects.

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+  foldSyntax go bound free = Scope . go (bound . fmap (go free)) . unScope
+
+instance (Syntax l, Syntax r) => Syntax (l :+: r) where
+  foldSyntax go bound free (L l) = L (foldSyntax go bound free l)
+  foldSyntax go bound free (R r) = R (foldSyntax go bound free r)
+
+
+iter :: forall m n sig a b
+     .  Syntax sig
+     => (forall a . m a -> n a)
```
robrix: Interpret free variables…

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+     -> (forall a . sig n a -> n a)
```
robrix: …and syntax…

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+     -> (forall a . Incr () (n a) -> m (Incr () (n a)))
```
robrix: …replacing bound variables and modifying scoped terms…

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+     -> (a -> m b)
```
robrix: …and mapping the initial free variables with a continuation (to avoid quadratic `fmap`s)…

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+     -> Term sig a
```
robrix: …to turn a `Term`…

semantic-core/src/Data/Term.hs
```diff
@@ -0,0 +1,93 @@
...
+     -> n b
```
robrix: …into an `n`.

semantic-core/test/Generators.hs
```diff
@@ -18,39 +18,40 @@ import qualified Hedgehog.Range as Range
...
 import Data.Core
 import Data.Name
+import Data.Term
 
 -- The 'prune' call here ensures that we don't spend all our time just generating
 -- fresh names for variables, since the length of variable names is not an
 -- interesting property as they parse regardless.
-name :: MonadGen m => m Name
-name = Gen.prune (User <$> names) where
+name :: MonadGen m => m (Named User)
```
robrix: Parsed/pretty-printed `Core` should _never_ contain `Gensym`s, so we simply disallow them by using `User` instead of `Name`.

PR #197, opened 2019-07-12
Protoc rework

tclem:
Makes the process of generating haskell code from proto files a bit more repeatable by doing all the work in a docker container instead of relying on people's local setup of the protocol buffer complier and associated plugins. There are a couple of rough edges here due to https://github.com/tclem/twirp-haskell not being quite powerful enough, but it gets the job done. I'm also moving the service proto definitions out of the repo as semantic doesn't expose a service.


Dockerfile
```diff
@@ -1,3 +1,18 @@
+# Put protoc and twirp tooling in it's own image
```
robrix: Nitpick:

```suggestion
# Put protoc and twirp tooling in its own image
```

PR #200, opened 2019-07-18
Fold over Core without iter

robrix:
This PR explores folds over `Core` without `iter`—direct recursion, `fromScope`, and so on.

I’m sorry to be removing `iter` and `cata`, but recursion schemes over nested datatypes require a bunch of extra machinery that I’d rather avoid.

- [x] Depends on #184.


semantic-core/src/Data/Core.hs
```diff
@@ -194,25 +194,8 @@ annWith :: (Carrier sig m, Member Core sig) => CallStack -> m a -> m a
...
-    Edge e t -> Edge e (go h t)
-    Frame -> Frame
-    a :. b -> go h a :. go h b
-    a := b -> go h a := go h b
-    Ann loc t -> Ann loc (go h t)
+stripAnnotations :: (Member Core sig, HFunctor sig, forall g . Functor g => Functor (sig g)) => Term sig a -> Term sig a
+stripAnnotations (Var v)  = Var v
+stripAnnotations (Term t)
+  | Just c <- prj t, Ann _ b <- c = stripAnnotations b
+  | otherwise                     = Term (hmap stripAnnotations t)
```
robrix: This is slightly less pretty than the `iter` definition, but is equivalent in effect and may even be a little faster (tho I certainly haven’t measured).

semantic-core/src/Data/Core.hs
```diff
@@ -194,25 +194,8 @@ annWith :: (Carrier sig m, Member Core sig) => CallStack -> m a -> m a
 annWith callStack = maybe id (fmap send . Ann) (stackLoc callStack)
 
 
-stripAnnotations :: (Member Core sig, Syntax sig) => Term sig a -> Term sig a
-stripAnnotations = iter id alg Var Var
-  where alg t | Just c <- prj t, Ann _ b <- c = b
-              | otherwise                     = Term t
-
-
-instance Syntax Core where
```
robrix: Goodbye!

semantic-core/src/Data/Core/Pretty.hs
```diff
@@ -56,57 +55,58 @@ inParens amount go = do
...
-    f <- lhs
-    g <- rhs
-    pure (f <+> symbol "=" <+> g)
-
-  -- Annotations are not pretty-printed, as it lowers the signal/noise ratio too profoundly.
-  Ann _ (Const c) -> c
-  where bind (Ignored x) f = let x' = name x in (,) x' <$> local (x':) (getConst (unScope f))
+prettyCore :: Style -> Term Core User -> AnsiDoc
+prettyCore style = run . runReader @Prec 0 . go (pure . name)
+  where go :: (Member (Reader Prec) sig, Carrier sig m) => (a -> m AnsiDoc) -> Term Core a -> m AnsiDoc
```
robrix: `go` is a direct-recursive worker function, walking the `Term` via a helper continuation which we extend to deal with `Incr`s efficiently as we walk under binders.

semantic-core/src/Data/Core/Pretty.hs
```diff
@@ -56,57 +55,58 @@ inParens amount go = do
...
+              pure (f <> symbol "." <> g)
+
+            lhs := rhs -> inParens 3 $ do
+              f <- go var lhs
+              g <- go var rhs
+              pure (f <+> symbol "=" <+> g)
+
+            -- Annotations are not pretty-printed, as it lowers the signal/noise ratio too profoundly.
+            Ann _ c -> go var c
+          where bind (Ignored x) f = let x' = name x in (,) x' <$> go (incr (const (pure x')) var) (fromScope f)
```
robrix: This is the real meat of the change: we use `fromScope` to transform the `Scope () f a` (which is isomorphic to `f (Incr () (f a))`) to `f (Incr () a)`, and then call `go` to recur over it with an extended continuation. I.e. if `var :: a -> m AnsiDoc`, then `incr (const (pure x')) var :: Incr () a -> m AnsiDoc`. Using this continuation enables us to avoid performing _n²_ `fmap`s as we recur through the structure.

semantic-core/src/Data/Term.hs
```diff
@@ -48,46 +41,3 @@ instance RightModule sig => Monad (Term sig) where
...
-     -> (forall a . sig (Const b) a -> b)
-     -> (Incr () b -> a)
-     -> (x -> a)
-     -> Term sig x
-     -> b
-cata var alg k h = getConst . iter (coerce var) (Const . alg) (coerce k) (Const . h)
-
-
-interpret :: (Carrier sig m, Member eff sig, Syntax eff) => (forall a . Incr () (m a) -> m (Incr () (m a))) -> (a -> m b) -> Term eff a -> m b
-interpret = iter id send
```
robrix: If we want to implement something like `interpret` again in the future, it should be easy enough since it shouldn’t actually require a `Syntax` instance to walk the `Term` direct-recursively, just an `HFunctor` instance.

PR #201, opened 2019-07-18
Types as syntax

robrix:
This PR reframes `Monotype` and `Polytype` as scope-safe syntax in `Term`.

- [x] Depends on #184.
- [x] Depends on #200.


semantic-core/src/Analysis/Typecheck.hs
```diff
@@ -29,102 +27,84 @@ import qualified Data.IntSet as IntSet
...
-        go i (PArr a b)             = PArr (go i a) (go i b)
-        go i (PRecord fs)           = PRecord (go i <$> fs)
-
-
-typecheckingFlowInsensitive :: [File (Term Core.Core Name)] -> (Heap Name (Monotype Meta), [File (Either (Loc, String) Polytype)])
+generalize :: Term Monotype Meta -> Term (Polytype :+: Monotype) Void
+generalize ty = fromJust (closed (forAlls (IntSet.toList (mvs ty)) (hoistTerm R ty)))
+
+
+typecheckingFlowInsensitive :: [File (Term Core.Core Name)] -> (Heap Name (Term Monotype Meta), [File (Either (Loc, String) (Term (Polytype :+: Monotype) Void))])
```
robrix: The types here are super interesting. Typechecking returns a polytype…

semantic-core/src/Analysis/Typecheck.hs
```diff
@@ -29,102 +27,84 @@ import qualified Data.IntSet as IntSet
...
 typecheckingFlowInsensitive
   = run
   . runFresh
   . runNaming
   . runHeap (Gen (Gensym (Nil :> "root") 0))
-  . (>>= traverse (traverse (traverse generalize)))
+  . fmap (fmap (fmap (fmap generalize)))
```
robrix: …which is the result of generalizing (quantifying over all of the variables in)…

semantic-core/src/Analysis/Typecheck.hs
```diff
@@ -29,102 +27,84 @@ import qualified Data.IntSet as IntSet
...
 runFile :: ( Carrier sig m
            , Effect sig
            , Member Fresh sig
            , Member Naming sig
-           , Member (State (Heap Name (Monotype Meta))) sig
+           , Member (State (Heap Name (Term Monotype Meta))) sig
            )
         => File (Term Core.Core Name)
-        -> m (File (Either (Loc, String) (Monotype Meta)))
+        -> m (File (Either (Loc, String) (Term Monotype Meta)))
```
robrix: …the monotype produced by `runFile`. Classic Hindley-Milner type schemery!

PR #212, opened 2019-07-31
New options for readBlobsFromGitRepo

tclem:
Allows specifying only the paths you want to parse when reading from a git repo. This is the inverse of the existing args `--exclude` and `--exclude-stdin`.

### DIFF
--- a/src/Data/Blob/IO.hs
+++ b/src/Data/Blob/IO.hs
@@ -39,8 +39,8 @@ readBlobsFromDir path = liftIO . fmap catMaybes $
   findFilesInDir path supportedExts mempty >>= Async.mapConcurrently (readBlobFromFile . fileForPath)
 
 -- | Read all blobs from the Git repo with Language.supportedExts
-readBlobsFromGitRepo :: MonadIO m => FilePath -> Git.OID -> [FilePath] -> m [Blob]
-readBlobsFromGitRepo path oid excludePaths = liftIO . fmap catMaybes $
+readBlobsFromGitRepo :: MonadIO m => FilePath -> Git.OID -> [FilePath] -> [FilePath] -> m [Blob]
+readBlobsFromGitRepo path oid excludePaths includePaths = liftIO . fmap catMaybes $
   Git.lsTree path oid >>= Async.mapConcurrently (blobFromTreeEntry path)
   where
     -- Only read tree entries that are normal mode, non-minified blobs in a language we can parse.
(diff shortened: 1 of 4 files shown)


src/Semantic/CLI.hs
```diff
@@ -114,7 +114,9 @@ parseCommand = command "parse" (info parseArgumentsParser (progDesc "Generate pa
                       <$> option str (long "gitDir" <> help "A .git directory to read from")
                       <*> option shaReader (long "sha" <> help "The commit SHA1 to read from")
                       <*> ( ExcludePaths <$> many (option str (long "exclude" <> short 'x' <> help "Paths to exclude"))
-                        <|> ExcludeFromHandle <$> flag' stdin (long "exclude-stdin" <> help "Exclude paths given to stdin"))
+                        <|> ExcludeFromHandle <$> flag' stdin (long "exclude-stdin" <> help "Exclude paths given to stdin")
+                        <|> OnlyPaths <$> many (option str (long "only" <> help "Only include the specified paths"))
+                        <|> OnlyPathsFromHandle <$> flag' stdin (long "only-stdin" <> help "Include only the paths given to stdin"))
```
robrix: The symmetry centre of my brain suggests: perhaps `IncludePaths` and `IncludeFromHandle` or `IncludePathsFromHandle` to mirror the exclude ones? I’m not super fussed, tho.

src/Data/Blob/IO.hs
```diff
@@ -50,6 +50,7 @@ readBlobsFromGitRepo path oid excludePaths = liftIO . fmap catMaybes $
       , lang `elem` codeNavLanguages
       , not (pathIsMinified path)
       , path `notElem` excludePaths
+      , null includePaths || path `elem` includePaths
```
robrix: I have some mild discomfort around “what does it mean if we have the same path in both `excludePaths` and `includePaths`, and will this be surprising?” but since the CLI doesn’t allow you to specify both I think that’s ok 👍

tclem: Ah yeah, if you call this function directly and pass the same path in both the excludes and includes list then excludes always win out. I think this might actually be fine/expected behavior.

src/Semantic/CLI.hs
tclem: I like that. I'm going to leave the command line switches because I think `--only` is a better description of the behavior, but I can rename the type constructors.

PR #213, opened 2019-07-31
Git ls tree fixes

tclem:
In the switch to autoparsec parsing of `git ls-tree` output (see #99) we lost the ability to handle things like commit objects (used for submodules) and other valid modes. Unfortunately things fail by just returning an empty list if parsing fails so this was a little challenging to catch.

This fixes the parsing and adds a test.

### DIFF
--- a/semantic.cabal
+++ b/semantic.cabal
@@ -1,7 +1,7 @@
 cabal-version:       2.4
 
 name:                semantic
-version:             0.6.0.0
+version:             0.7.0.0
 synopsis:            Framework and executable for analyzing and diffing untrusted code.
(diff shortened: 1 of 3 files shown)


semantic.cabal
```diff
@@ -1,7 +1,7 @@
 cabal-version:       2.4
 
 name:                semantic
-version:             0.6.0.0
+version:             0.7.0.0
```
robrix: This appears to be a backwards-compatible bugfix, so I think this could be 0.6.0.1?

tclem: I meant to bump this on #212 since that added some new cmd line args, but didn't push the commit before that PR merged.

PR #229, opened 2019-08-14
Use TSX parser instead of Typescript parser for JavaScript and JSX

maxbrunsfeld:
JavaScript (and `.jsx`) files can contain JSX (and cannot use TypeScript's type assertion syntax), so we need to use the TSX dialect of TypeScript when parsing and analyzing them.


maxbrunsfeld:
Is there any testing that should be added for this? Previously, we would have gotten parse errors on any `.js` files that used JSX, so maybe just a unit test that we can parse a `.js` file with some JSX in it?

tclem:
Let's do something like [this](https://github.com/github/semantic/blob/a6c073895cae7718113197ca0c8db2b2b7286f6f/test/fixtures/tsx/corpus/jsx-elements.A.tsx), but with a `.jsx` extension. You can use `script/generate-example` to help bootstrap the expected output.

maxbrunsfeld:
While we were here, @rewinfrey and I also added some limited support for Type arguments in TSX (https://github.com/Microsoft/TypeScript/pull/22415).

test/fixtures/tsx/corpus/jsx-elements.diffA-B.txt
```diff
@@ -6,6 +6,7 @@
       (JsxElement
         (JsxOpeningElement
           (Identifier)
+          (Empty)
```
maxbrunsfeld: `JsxOpeningElement` terms now have an optional `TypeArguments` child.

erickadalberto:
> JavaScript files can contain JSX (and cannot use TypeScriptyp so we need to use *the TSX dialect of TypeScript when parsing and analyzing them\

PR #252, opened 2019-09-16
New version of lingo-haskell

tclem:
++lingo-haskell to pick up https://github.com/tclem/lingo-haskell/pull/2.


robrix:
I don’t understand how the tests can be failing given that this doesn’t touch the `tree-sitter-*` packages 😳

PR #265, opened 2019-09-20
Fixed Range Ord instance

robrix:
I recently discovered [a long-standing bug](https://github.com/github/semantic/commit/44747394271d7e504b1391e6d11329d0b19bf79c): `Range`’s `Ord` instance doesn’t consider the `end`, only the `start`, which means that `Range 0 10` and `Range 0 0` are ordered as `EQ`, which is both lawless and a bad idea.

This behaviour has existed for just shy of four years, so I have some concern that it might be a load-bearing bug (i.e. that we might have other behaviour dependent on this misbehaviour). However, I believe the `Ord` instance is unused except for caching analyses (which should probably be ignoring the annotations anyway), so 🤞

I’ve also taken the opportunity to remove some unused functions.

### DIFF
--- a/src/Data/Range.hs
+++ b/src/Data/Range.hs
@@ -15,7 +15,7 @@ import Data.JSON.Fields
 
 -- | A half-open interval of integers, defined by start & end indices.
 data Range = Range { start :: {-# UNPACK #-} !Int, end :: {-# UNPACK #-} !Int }
-  deriving (Eq, Generic, NFData)
+  deriving (Eq, Generic, NFData, Ord)
 
 emptyRange :: Range
 emptyRange = Range 0 0
@@ -41,9 +41,6 @@ subtractRange range1 range2 = Range (start range1) (end range1 - rangeLength (Ra
 -- | The associativity of this instance is specced in @Data.Range.Spec@.
 instance Semigroup Range where
   Range start1 end1 <> Range start2 end2 = Range (min start1 start2) (max end1 end2)
-
-instance Ord Range where
-  a <= b = start a <= start b
 
 instance Show Range where
   showsPrec _ Range{..} = showChar '[' . shows start . showString " .. " . shows end . showChar ']'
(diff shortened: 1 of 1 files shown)


src/Data/Range.hs
```diff
@@ -24,14 +22,6 @@ emptyRange = Range 0 0
 rangeLength :: Range -> Int
 rangeLength range = end range - start range
 
--- | Offset a range by a constant delta.
-offsetRange :: Range -> Int -> Range
-offsetRange a b = Range (start a + b) (end a + b)
-
--- | Test two ranges for intersection.
-intersectsRange :: Range -> Range -> Bool
-intersectsRange range1 range2 = start range1 < end range2 && start range2 < end range1
```
robrix: We weren’t using either of these anywhere.

src/Data/Range.hs
```diff
@@ -42,9 +32,6 @@ subtractRange range1 range2 = Range (start range1) (end range1 - rangeLength (Ra
 instance Semigroup Range where
   Range start1 end1 <> Range start2 end2 = Range (min start1 start2) (max end1 end2)
 
-instance Ord Range where
-  a <= b = start a <= start b
```
robrix: As far as I can recall, I defined this in the first place because I didn’t understand either a) the rules that derived `Ord` instances obey (especially w.r.t. lexicographical ordering), or b) the laws governing the expected behaviour of `Ord` instances in general.

One such law is that `Ord` instances should agree with `Eq` instances (`Eq` is a superclass of `Ord`, after all), which this instance violated: `x <= y && y <= x = x == y` doesn’t hold; equivalently, `compare x y == EQ = x == y` doesn’t hold.

robrix:
It typechecks!

The tests pass!

…just like they did when we had the bug for four years.

mbj:
> …just like they did when we had the bug for four years.

Could liquid Haskell have helped to spot the parent typeclass law violation? It to my naive eyes looks like a violation an SMT solver should easily be able to find.

robrix:
@mbj I’m no expert, but that does sound like a good job for liquid haskell.

mbj:
> I’m no expert, but that does sound like a good job for liquid haskell.

No expert here either, but wanted to confirm my mental model. Thx for doing so.

PR #269, opened 2019-09-20
Factor source code-related facilities into a new package

robrix:
This PR factors `Source`, `Range`, `Span`, `Location`, &c. out into a new package, `semantic-source`, intended to be used by `semantic`, `semantic-ast`, `semantic-tags`, `semantic-core`, `semantic-python`, and `haskell-tree-sitter`.


semantic-source/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2015-2019 GitHub
```
robrix: A lot of this code has been sitting around since 2015. (At a minimum, the `Range` stuff.)

semantic-source/semantic-source.cabal
```diff
@@ -0,0 +1,88 @@
...
+    -Wno-all-missed-specialisations
+    -Wno-star-is-type
+
+library
+  import: common
+  exposed-modules:
+    Source.Loc
+    Source.Range
+    Source.Source
+    Source.Span
```
robrix: IMO we should format `.cabal` files this way in general. Aligning crap to the right of labels is terrible.

semantic-source/src/Source/Loc.hs
```diff
@@ -0,0 +1,40 @@
...
+
+import Control.DeepSeq (NFData)
+import Data.Hashable (Hashable)
+import Data.Monoid.Generic
+import GHC.Generics (Generic)
+import Prelude hiding (span)
+import Source.Range
+import Source.Span
+
+data Loc = Loc
```
robrix: I opted to abbreviate `Location` to `Loc` because we use it _everywhere_.

semantic-source/src/Source/Loc.hs
```diff
@@ -0,0 +1,40 @@
...
+  { byteRange :: {-# UNPACK #-} !Range
+  , span      :: {-# UNPACK #-} !Span
```
robrix: Furthermore, and in keeping with our style guide, I’ve opted to rename its fields to drop the prefix.

semantic-source/src/Source/Loc.hs
```diff
@@ -0,0 +1,40 @@
...
+instance Hashable Loc
+instance NFData   Loc
+
+instance HasSpan Loc where
+  span_ = lens span (\l s -> l { span = s })
+  {-# INLINE span_ #-}
+
+
+byteRange_ :: Lens' Loc Range
+byteRange_ = lens byteRange (\l r -> l { byteRange = r })
```
robrix: I also gave it a handy lens to modify its range. We could do a classy-lenses–style `HasRange` eventually but for now this is it.

semantic-source/src/Source/Range.hs
```diff
@@ -0,0 +1,62 @@
...
+instance Semigroup Range where
+  Range start1 end1 <> Range start2 end2 = Range (min start1 start2) (max end1 end2)
+
+instance Lower Range where
+  lowerBound = Range 0 0
+
+
+-- | Construct a 'Range' with a given value for both its start and end indices.
+point :: Int -> Range
+point i = Range i i
```
robrix: This is especially useful alongside the `Semigroup` instance.

semantic-source/src/Source/Source.hs
```diff
@@ -0,0 +1,142 @@
+{-# LANGUAGE DeriveGeneric, GeneralizedNewtypeDeriving #-}
+{-|
+'Source' models source code, represented as a thin wrapper around a 'B.ByteString' with conveniences for splitting by line, slicing, etc.
+
+This module is intended to be imported qualified to avoid name clashes with 'Prelude':
```
robrix: We dropped prefixes from this module (🎩 @patrickt!) in keeping with the style guide.

src/Analysis/PackageDef.hs
```diff
@@ -50,13 +50,13 @@ instance (PackageDefStrategy syntax ~ strategy, HasPackageDefWithStrategy strate
...
   -- | Produce a customized 'PackageDef' for a given syntax node.
-  customToPackageDef :: (Foldable whole) => Blob -> Location -> syntax (Term whole Location, Maybe PackageDef) -> Maybe PackageDef
+  customToPackageDef :: (Foldable whole) => Blob -> Loc -> syntax (Term whole Loc, Maybe PackageDef) -> Maybe PackageDef
 
 
 instance CustomHasPackageDef Language.Go.Syntax.Package where
   customToPackageDef Blob{..} _ (Language.Go.Syntax.Package (Term (In fromAnn _), _) _)
     = Just $ PackageDef (getSource fromAnn)
-    where getSource = toText . flip Source.slice blobSource . locationByteRange
+    where getSource = toText . Source.slice blobSource . byteRange
```
robrix: We flipped `Source.slice`’s definition since we so frequently had to flip its uses instead.

src/Control/Abstract/Heap.hs
```diff
@@ -180,7 +180,7 @@ define :: ( HasCallStack
        -> Evaluator term address value m ()
 define declaration rel accessControl def = withCurrentCallStack callStack $ do
   -- TODO: This span is still wrong.
-  declare declaration rel accessControl emptySpan Unknown Nothing
+  declare declaration rel accessControl lowerBound Unknown Nothing
```
robrix: `emptySpan` is gone; we rely on the `Lower` instance instead.

src/Data/Abstract/ScopeGraph.hs
```diff
@@ -103,8 +103,8 @@ data Info scopeAddress = Info
   } deriving (Eq, Show, Ord, Generic, NFData)
 
 instance HasSpan (Info scopeAddress) where
-  span = lens infoSpan (\i s -> i { infoSpan = s })
-  {-# INLINE span #-}
+  span_ = lens infoSpan (\i s -> i { infoSpan = s })
+  {-# INLINE span_ #-}
```
robrix: We renamed the `Span` fields to `start` and `end`, and now follow the convention that lens names end in `_`.

src/Data/JSON/Fields.hs
```diff
@@ -46,6 +48,15 @@ instance Apply ToJSONFields1 fs => ToJSONFields1 (Sum fs) where
...
   toJSONFields (a, b) = [ "before" .= JSONFields a, "after" .= JSONFields b ]
 
+instance ToJSONFields Range where
+  toJSONFields Range{..} = ["sourceRange" .= [ start, end ]]
+
+instance ToJSONFields Span where
+  toJSONFields sourceSpan = [ "sourceSpan" .= sourceSpan ]
+
+instance ToJSONFields Loc where
+  toJSONFields Loc{..} = toJSONFields byteRange <> toJSONFields span
```
robrix: All these instances live here now so we didn’t have to move `ToJSONFields` into `semantic-source` or anything.

src/Parsing/CMark.hs
```diff
@@ -51,43 +52,53 @@ exts = [
...
+
+
+spanToRangeInLineRanges :: Array Int Range -> Span -> Range
+spanToRangeInLineRanges lineRanges Span{..} = Range
+  (Range.start (lineRanges ! line start) + pred (column start))
+  (Range.start (lineRanges ! line end)   + pred (column end))
+
+sourceLineRangesByLineNumber :: Source -> Array Int Range
+sourceLineRangesByLineNumber source = listArray (1, length lineRanges) lineRanges
+  where lineRanges = Source.lineRanges source
```
robrix: These were only ever being used in this module, and so moving them here prevented having to have `semantic-source` depend on `array`.

robrix:
One last bit: we’ve opted to leave `Blob` in `semantic`, for now.

robrix:
https://hackage.haskell.org/package/semantic-source-0.0.0.0

PR #295, opened 2019-09-30
Parse effect

robrix:
This PR:

- [x] Factors a `Parse` effect out of `Task`.
- [x] Defines two separate carriers for `Parse`: one which does all the telemetry stuff, and one which does not and therefore makes far fewer requests of its calling environment, making it more suitable for `ghci` & the tests.
- [x] Factors the rest of the actions, all of which amounted to pure functions, out of `Task` as well.
- [x] 🔥s `Task`.

Two things I’d like to try with this but haven’t got that far yet:

1. Make the `Parser` constructors into separate constructors of the `Parse` effect instead, so instead of doing `parse pythonParser blob` you’d have `parsePython blob` or something like that.

2. Define the `Measured` carrier for the `Parse` effect by composing onto the basic one, just layering in the timing/telemetry/etc.

### DIFF
--- a/src/Control/Carrier/Parse/Measured.hs
+++ b/src/Control/Carrier/Parse/Measured.hs
@@ -1,5 +1,5 @@
 {-# LANGUAGE GADTs, GeneralizedNewtypeDeriving, TypeOperators, UndecidableInstances #-}
-module Semantic.Parse
+module Control.Carrier.Parse.Measured
 ( -- * Parse effect
   module Control.Effect.Parse
   -- * Parse carrier
--- a/src/Control/Carrier/Parse/Measured.hs
+++ b/src/Control/Carrier/Parse/Measured.hs
@@ -59,13 +59,13 @@ runParser blob@Blob{..} parser = case parser of
     time "parse.tree_sitter_ast_parse" languageTag $ do
       config <- asks config
       parseToAST (configTreeSitterParseTimeout config) language blob
-        >>= maybeM (throwError (SomeException ParserTimedOut))
+        >>= either (trace >=> const (throwError (SomeException ParserTimedOut))) pure
 
   UnmarshalParser language ->
     time "parse.tree_sitter_ast_parse" languageTag $ do
       config <- asks config
       parseToPreciseAST (configTreeSitterParseTimeout config) language blob
-        >>= maybeM (throwError (SomeException ParserTimedOut))
+        >>= either (trace >=> const (throwError (SomeException ParserTimedOut))) pure
 
   AssignmentParser    parser assignment -> runAssignment Assignment.assign    parser blob assignment
   DeterministicParser parser assignment -> runAssignment Deterministic.assign parser blob assignment
--- a/src/Control/Carrier/Parse/Measured.hs
+++ b/src/Control/Carrier/Parse/Measured.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE GADTs, GeneralizedNewtypeDeriving, TypeOperators, UndecidableInstances #-}
+-- | A carrier for 'Parse' effects suitable for use in production.
 module Control.Carrier.Parse.Measured
 ( -- * Parse effect
   module Control.Effect.Parse
--- a/src/Control/Carrier/Parse/Simple.hs
+++ b/src/Control/Carrier/Parse/Simple.hs
@@ -4,6 +4,7 @@ module Control.Carrier.Parse.Simple
   module Control.Effect.Parse
   -- * Parse carrier
 , ParseC(..)
+, runParse
   -- * Exceptions
 , ParserCancelled(..)
 ) where
@@ -19,8 +20,6 @@ import           Control.Exception
 import           Control.Monad.IO.Class
 import           Data.Blob
 import qualified Data.Error as Error
-import qualified Data.Flag as Flag
-import qualified Data.Syntax as Syntax
 import           Data.Sum
 import           Data.Term
 import           Data.Typeable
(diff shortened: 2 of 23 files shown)


src/Control/Carrier/Parse/Measured.hs
```diff
@@ -0,0 +1,160 @@
+{-# LANGUAGE GADTs, GeneralizedNewtypeDeriving, TypeOperators, UndecidableInstances #-}
+-- | A carrier for 'Parse' effects suitable for use in production.
+module Control.Carrier.Parse.Measured
```
robrix: This module’s contents were mostly just moved here from `Semantic.Task`.

src/Control/Carrier/Parse/Measured.hs
```diff
@@ -0,0 +1,160 @@
...
+         , Member Timeout sig
+         , Member Trace sig
+         , MonadIO m
+         )
+      => Carrier (Parse :+: sig) (ParseC m) where
+  eff (L (Parse parser blob k)) = runParser blob parser >>= k
+  eff (R other) = ParseC (eff (handleCoercible other))
+
+-- | Parse a 'Blob' in 'IO'.
+runParser :: (Member (Error SomeException) sig, Member (Reader TaskSession) sig, Member Telemetry sig, Member Timeout sig, Member Trace sig, Carrier sig m, MonadIO m)
```
robrix: In particular, the only changes I made to `runParser`/`runAssignment`/etc were to remove a couple of redundant `Member (Lift IO) sig` constraints.

src/Control/Carrier/Parse/Simple.hs
```diff
@@ -0,0 +1,73 @@
+{-# LANGUAGE GADTs, GeneralizedNewtypeDeriving, TypeOperators, UndecidableInstances #-}
+-- | A carrier for 'Parse' effects suitable for use in the repl, tests, etc.
+module Control.Carrier.Parse.Simple
```
robrix: By contrast, I wrote this module by starting with the other one and then paring it down significantly.

src/Control/Effect/Parse.hs
```diff
@@ -0,0 +1,29 @@
+{-# LANGUAGE ExistentialQuantification #-}
+module Control.Effect.Parse
+( -- * Parse effect
+  Parse(..)
+, parse
```
robrix: By factoring this out into its own effect, stuff that needs to parse no longer needs most of the other junk that `Task` came with, and thus can be simplified overall.

src/Parsing/TreeSitter.hs
```diff
@@ -33,43 +32,36 @@ import qualified TreeSitter.Unmarshal as TS
...
-              , Carrier sig m
               , Enum grammar
-              , Member Trace sig
               , MonadIO m
               )
            => Duration
            -> Ptr TS.Language
            -> Blob
-           -> m (Maybe (AST [] grammar))
+           -> m (Either String (AST [] grammar))
```
robrix: Returning `Either` allowed us to drop the `Member Trace sig` constraint.

src/Semantic/AST.hs
```diff
@@ -35,10 +38,10 @@ astParseBlob blob@Blob{..}
 data ASTFormat = SExpression | JSON | Show | Quiet
   deriving (Show)
 
-runASTParse :: (Member Distribute sig, Member (Error SomeException) sig, Member Task sig, Carrier sig m, MonadIO m) => ASTFormat -> [Blob] -> m F.Builder
+runASTParse :: (Member Distribute sig, Member (Error SomeException) sig, Member Parse sig, Member (Reader Config) sig, Carrier sig m, MonadIO m) => ASTFormat -> [Blob] -> m F.Builder
```
robrix: Serialization has been factored out, and needs the `Config` to do the colour/plain output switch.

src/Semantic/AST.hs
```diff
@@ -35,10 +38,10 @@ astParseBlob blob@Blob{..}
...
 runASTParse SExpression = distributeFoldMap (astParseBlob >=> withSomeAST (serialize (F.SExpression F.ByShow)))
 runASTParse Show        = distributeFoldMap (astParseBlob >=> withSomeAST (serialize F.Show . fmap nodeSymbol))
-runASTParse JSON        = distributeFoldMap (\ blob -> astParseBlob blob >>= withSomeAST (render (renderJSONAST blob))) >=> serialize F.JSON
+runASTParse JSON        = distributeFoldMap (\ blob -> withSomeAST (renderJSONAST blob) <$> astParseBlob blob) >=> serialize F.JSON
```
robrix: Rendering is now pure.

src/Semantic/Api/Diffs.hs
```diff
@@ -85,19 +89,19 @@ diffGraph blobs = DiffTreeGraphResponse . V.fromList . toList <$> distributeFor
...
-dotGraphDiff blobPair = doDiff blobPair (const pure) render
+dotGraphDiff blobPair = doDiff blobPair (const id) render
   where render _ = serialize (DOT (diffStyle "diffs")) . renderTreeGraph
 
-type DiffEffects sig m = (Member (Error SomeException) sig, Member Telemetry sig, Member Distribute sig, Member Task sig, Carrier sig m, MonadIO m)
+type DiffEffects sig m = (Member (Error SomeException) sig, Member (Reader Config) sig, Member Telemetry sig, Member Distribute sig, Member Parse sig, Carrier sig m, MonadIO m)
 
 type CanDiff syntax = (ConstructorName syntax, Diffable syntax, Eq1 syntax, HasDeclaration syntax, Hashable1 syntax, Show1 syntax, ToJSONFields1 syntax, Traversable syntax)
-type Decorate m a b = forall syntax . CanDiff syntax => Blob -> Term syntax a -> m (Term syntax b)
+type Decorate a b = forall syntax . CanDiff syntax => Blob -> Term syntax a -> Term syntax b
```
robrix: Decorating is pure.

src/Semantic/Api/Diffs.hs
```diff
@@ -111,33 +115,33 @@ type TermPairConstraints =
...
   SomeTermPair terms <- doParse blobPair decorate
   diff <- diffTerms blobPair terms
   render blobPair diff
 
-diffTerms :: (CanDiff syntax, Member Task sig, Member Telemetry sig, Carrier sig m, MonadIO m)
+diffTerms :: (CanDiff syntax, Member Telemetry sig, Carrier sig m, MonadIO m)
   => BlobPair -> Join These (Term syntax ann) -> m (Diff syntax ann ann)
 diffTerms blobs terms = time "diff" languageTag $ do
-  diff <- diff (runJoin terms)
+  let diff = diffTermPair (runJoin terms)
```
robrix: Diffing is pure.

src/Semantic/CLI.hs
```diff
@@ -57,13 +58,13 @@ main = do
   (options, task) <- customExecParser (prefs showHelpOnEmpty) arguments
   config <- defaultConfig options
   res <- withTelemetry config $ \ (TelemetryQueues logger statter _) ->
-    Task.runTask (Task.TaskSession config "-" (optionsLogPathsOnError options) logger statter) task
+    Task.runTask (Task.TaskSession config "-" (optionsLogPathsOnError options) logger statter) (Parse.runParse task)
```
robrix: `runTask` doesn’t run the parser, for two reasons:

1. A lot of the time when we do `runTask`, we’re only doing it to get all the effects that the `Measured` parse carrier requires; by splitting them up, it’ll be a bit easier to start doing away with `runTask` altogether.

2. I didn’t want to deal with circular imports between the measured carrier and `Semantic.Task` to deal with the `TaskSession`, so `Semantic.Task` doesn’t import the measured parse carrier.

src/Semantic/Graph.hs
```diff
@@ -230,7 +231,7 @@ runScopeGraph :: Ord address
 runScopeGraph = raiseHandler (runState lowerBound)
 
 -- | Parse a list of files into a 'Package'.
-parsePackage :: (Member Distribute sig, Member (Error SomeException) sig, Member Resolution sig, Member Task sig, Member Trace sig, Carrier sig m)
+parsePackage :: (Member Distribute sig, Member (Error SomeException) sig, Member Resolution sig, Member Parse sig, Member Trace sig, Carrier sig m)
```
robrix: This still requires a whole whack of effects, but _not_ telemetry, the `TaskSession`, etc., or any of the other bits implied by `Task` but not necessarily implied by `Parse`.

src/Semantic/REPL.hs
```diff
@@ -57,19 +51,15 @@ instance Exception Quit
...
+  withOptions debugOptions $ \config _ _ ->
     runM
     . withDistribute
     . runCatch
-    . runResource
-    . withTimeout
     . runError @SomeException
-    . runTelemetryIgnoringStat (logOptionsFromConfig config)
-    . runTraceInTelemetry
-    . runReader (TaskSession config "-" False logger statter)
```
robrix: Case in point, we are able to delete a bunch of stuff from `repl` now that parsing doesn’t require all of it.

src/Semantic/Task.hs
```diff
@@ -88,14 +64,12 @@ import qualified Semantic.Task.Files as Files
...
-import           Source.Source (Source)
 
 -- | A high-level task producing some result, e.g. parsing, diffing, rendering. 'Task's can also specify explicit concurrency via 'distribute', 'distributeFor', and 'distributeFoldMap'
-type TaskEff
-  = TaskC
-  ( ResolutionC
+type TaskC
+  = ResolutionC
   ( Files.FilesC
+  ( ReaderC Config
```
robrix: Passing this around separately to enable `serialize` not to depend on the entire `TaskSession`.

src/Semantic/Task.hs
```diff
@@ -88,14 +64,12 @@ import qualified Semantic.Task.Files as Files
...
   ( ReaderC TaskSession
```
robrix: However, enough other stuff _does_ depend on the entire `TaskSession` that we’ve got to keep this around still.

src/Semantic/Task.hs
```diff
@@ -106,41 +80,13 @@ type TaskEff
...
 
--- | A function to render terms or diffs.
-type Renderer i o = i -> o
-
--- | A task which parses a 'Blob' with the given 'Parser'.
-parse :: (Member Task sig, Carrier sig m)
-      => Parser term
-      -> Blob
-      -> m term
-parse parser blob = send (Parse parser blob pure)
```
robrix: Moved into its own effect.

src/Semantic/Task.hs
```diff
@@ -106,41 +80,13 @@ type TaskEff
...
-
--- | A task which decorates a 'Term' with values computed using the supplied 'RAlgebra' function.
-decorate :: (Functor f, Member Task sig, Carrier sig m)
-         => RAlgebra (TermF f Loc) (Term f Loc) field
-         -> Term f Loc
-         -> m (Term f field)
-decorate algebra term = send (Decorate algebra term pure)
```
robrix: Replaced with pure calls to `decoratorWithAlgebra`.

src/Semantic/Task.hs
```diff
@@ -106,41 +80,13 @@ type TaskEff
...
-
--- | A task which diffs a pair of terms using the supplied 'Differ' function.
-diff :: (Diffable syntax, Eq1 syntax, Hashable1 syntax, Traversable syntax, Member Task sig, Carrier sig m)
```
robrix: Replaced with pure calls to `diffTermPair`.

src/Semantic/Task.hs
```diff
@@ -106,41 +80,13 @@ type TaskEff
...
-     => These (Term syntax ann) (Term syntax ann)
-     -> m (Diff syntax ann ann)
-diff terms = send (Semantic.Task.Diff terms pure)
-
--- | A task which renders some input using the supplied 'Renderer' function.
-render :: (Member Task sig, Carrier sig m)
```
robrix: Replaced with pure function application in the approximately two places this was even used.

src/Semantic/Task.hs
```diff
@@ -106,41 +80,13 @@ type TaskEff
...
-
-serialize :: (Member Task sig, Carrier sig m)
+serialize :: (Member (Reader Config) sig, Carrier sig m)
           => Format input
           -> input
           -> m Builder
-serialize format input = send (Serialize format input pure)
+serialize format input = do
+  formatStyle <- asks (Flag.choose IsTerminal Plain Colourful . configIsTerminal)
+  pure (runSerialize formatStyle format input)
```
robrix: For this one I kept the function but just moved the handler code here.

PR #300, opened 2019-10-02
Abstract actions over terms

robrix:
This PR abstracts actions on terms and diffs over the term/diff types instead of the syntax type.

### DIFF
--- a/src/Semantic/Api/Diffs.hs
+++ b/src/Semantic/Api/Diffs.hs
@@ -62,7 +62,7 @@ parseDiffBuilder DiffDotGraph    = distributeFoldMap dotGraphDiff
 type RenderJSON m syntax = forall syntax . CanDiff syntax => BlobPair -> Diff syntax Loc Loc -> m (Rendering.JSON.JSON "diffs" SomeJSON)
 
 jsonDiff :: DiffEffects sig m => RenderJSON m syntax -> BlobPair -> m (Rendering.JSON.JSON "diffs" SomeJSON)
-jsonDiff f blobPair = doDiff (const id) f blobPair `catchError` jsonError blobPair
+jsonDiff f blobPair = doDiff (const id) (f blobPair) blobPair `catchError` jsonError blobPair
 
 jsonError :: Applicative m => BlobPair -> SomeException -> m (Rendering.JSON.JSON "diffs" SomeJSON)
 jsonError blobPair (SomeException e) = pure $ renderJSONDiffError blobPair (show e)
@@ -81,22 +81,23 @@ diffGraph blobs = DiffTreeGraphResponse . V.fromList . toList <$> distributeFor 
         path = T.pack $ pathForBlobPair blobPair
         lang = bridging # languageForBlobPair blobPair
 
-        render :: (Foldable syntax, Functor syntax, ConstructorName syntax, Applicative m) => BlobPair -> Diff syntax Loc Loc -> m DiffTreeFileGraph
-        render _ diff =
+        render :: (Foldable syntax, Functor syntax, ConstructorName syntax, Applicative m) => Diff syntax Loc Loc -> m DiffTreeFileGraph
+        render diff =
           let graph = renderTreeGraph diff
               toEdge (Edge (a, b)) = DiffTreeEdge (diffVertexId a) (diffVertexId b)
           in pure $ DiffTreeFileGraph path lang (V.fromList (vertexList graph)) (V.fromList (fmap toEdge (edgeList graph))) mempty
 
 
 sexpDiff :: DiffEffects sig m => BlobPair -> m Builder
-sexpDiff = doDiff (const id) (const (serialize (SExpression ByConstructorName)))
+sexpDiff = doDiff (const id) (serialize (SExpression ByConstructorName))
 
 showDiff :: DiffEffects sig m => BlobPair -> m Builder
-showDiff = doDiff (const id) (const (serialize Show))
+showDiff = doDiff (const id) (serialize Show)
 
 dotGraphDiff :: DiffEffects sig m => BlobPair -> m Builder
 dotGraphDiff = doDiff (const id) render
-  where render _ = serialize (DOT (diffStyle "diffs")) . renderTreeGraph
+  where render :: (Carrier sig m, ConstructorName syntax, Foldable syntax, Functor syntax, Member (Reader Config) sig) => Diff syntax Loc Loc -> m Builder
+        render = serialize (DOT (diffStyle "diffs")) . renderTreeGraph
 
 type DiffEffects sig m = (Member (Error SomeException) sig, Member (Reader Config) sig, Member Telemetry sig, Member Distribute sig, Member Parse sig, Carrier sig m, MonadIO m)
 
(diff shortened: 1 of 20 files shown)


src/Semantic/Api/Terms.hs
```diff
@@ -106,27 +84,71 @@ quietTerm blob = showTiming blob <$> time' ( (doParse blob >>= withSomeTerm (fma
...
+
+class (DOTGraphTerm t, JSONGraphTerm t, JSONTreeTerm t, SExprTerm t, ShowTerm t) => TermActions t
+instance (DOTGraphTerm t, JSONGraphTerm t, JSONTreeTerm t, SExprTerm t, ShowTerm t) => TermActions t
+
+doParse
+  :: ( Carrier sig m
+     , Member (Error SomeException) sig
+     , Member Parse sig
+     )
+  => (forall term . TermActions term => term Loc -> m a)
```
robrix: This function presents something of a problem, and I’m not happy with any of the solutions I’ve landed on thus far.

The problem: given _m_ languages, and _n_ features, fill in the matrix defined by their cartesian product incrementally; or, how do I add support for a language one feature at a time?

Diffing, parsing, analysis, and now symbols all have their own variations on this function, used to allow us to e.g. support `parse --symbols` for precise Python terms now, before we’ve added support for `parse --dot`, `diff`, or anything else. This is to some degree necessary: we don’t support analysis for Markdown or JSON, for example, because they have no computational content. Even so, it’s not without flaws:

- We have several different functions mapping languages to parsers, which increases our surface area for bugs; there’s no one _canonical_ mapping from languages to parsers. If we someday decide JavaScript should be parsed as TypeScript instead of TSX, we’ll have to be very careful to locate and update every place where we’re doing something `doParse`-esque.

- Each such function has to be typechecked independently, and that’s not quick—checking `TermActions` for all of these languages means solving all of its constraints for each `Term` instance, and therefore solving all of _their_ constraints, transitively, for every element of each syntax sum.

- Repeating that effort for each superclass of `TermActions` separately would require multiple `doParse` analogues, which we’d like to avoid due to the aforementioned duplication & margin for error. I have anecdotal evidence suggesting that it’s also slower than grouping them all up under `TermActions` thus.

- While we can easily parameterize `doParse` by the constraint we want to satisfy using `AllowAmbiguousTypes`, `ConstraintKinds`, & `TypeApplications`, doing so moves the burden of solving the constraints to each call-site, slowing down compilation significantly. It also requires us to explicitly mention all of the term types exhaustively in its context, which is tiresome to enumerate.

- Grouping multiple constraints into `TermActions` for improved compile times means that we can’t add support for each member constraint independently of the others; thus, we can’t add `--show` support for Python terms until we’ve also implemented `--dot`, `--json`, etc.

- To the best of my knowledge, there’s no way for us to turn the static satisfiability of a given constraint into a runtime `Maybe` (i.e. to return `Just` for any languages supporting the feature, `Nothing` for the others, using the same function for each different feature).

In an effort to rectify this situation, I have experimented with an approach to allow us to list the parsers for `ShowTerm` separately of the parsers for `JSONTreeTerm`, gaining the compile-time benefits of statically listing all of the supported languages for a given constraint, and the flexibility of tailoring these lists to each feature, without _quite_ so much duplication or compilation time. It’s still pretty duplicative of the map between languages and parsers, tho, and while it doesn’t seem to kill compile times straight away, more experience is needed to know whether it’s viable. I will be working on this in a follow-up PR.

src/Semantic/Api/Symbols.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE GADTs, TypeOperators, DerivingStrategies #-}
+{-# LANGUAGE MonoLocalBinds, RankNTypes #-}
```
patrickt: Out of curiosity, why `MonoLocalBinds` here?

src/Semantic/Api/Terms.hs
```diff
@@ -106,27 +84,70 @@ quietTerm blob = showTiming blob <$> time' ( (doParse blob >>= withSomeTerm (fma
...
-  JSX        -> SomeTerm <$> parse tsxParser blob
-  Markdown   -> SomeTerm <$> parse markdownParser blob
-  Python     -> SomeTerm <$> parse pythonParser blob
-  Ruby       -> SomeTerm <$> parse rubyParser blob
-  TypeScript -> SomeTerm <$> parse typescriptParser blob
-  TSX        -> SomeTerm <$> parse tsxParser blob
-  PHP        -> SomeTerm <$> parse phpParser blob
+
+class ShowTerm term where
+  showTerm :: (Carrier sig m, Member (Reader Config) sig) => term Loc -> m Builder
```
patrickt: Same thought on `Show` here? `SerializeTerm` might be better?

src/Semantic/Api/Terms.hs
patrickt: This is a wonderful and exhaustive analysis! Many thanks for it, and many thanks for caring about compile times. I don’t have any ideas on what the best way to do this is yet.

robrix: Fortunately, I’ve made a ton of headway here 👍

src/Semantic/Api/Diffs.hs
```diff
@@ -53,83 +56,139 @@ data DiffOutputFormat
...
+class JSONGraphDiff diff where
+  jsonGraphDiff :: BlobPair -> diff Loc Loc -> DiffTreeFileGraph
+
+instance (Foldable syntax, Functor syntax, ConstructorName syntax) => JSONGraphDiff (Diff syntax) where
+  jsonGraphDiff blobPair diff
+    = let graph = renderTreeGraph diff
+          toEdge (Edge (a, b)) = DiffTreeEdge (diffVertexId a) (diffVertexId b)
+      in DiffTreeFileGraph path lang (V.fromList (vertexList graph)) (V.fromList (fmap toEdge (edgeList graph))) mempty where
+        path = T.pack $ pathForBlobPair blobPair
+        lang = bridging # languageForBlobPair blobPair
```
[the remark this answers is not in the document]
robrix: This is copy-pasta, I’d rather not change it in this PR.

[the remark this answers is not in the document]
robrix: (Tho I agree.)

src/Semantic/Api/Diffs.hs
```diff
@@ -53,83 +56,139 @@ data DiffOutputFormat
...
-
-type RenderJSON m syntax = forall syntax . CanDiff syntax => BlobPair -> Diff syntax Loc Loc -> m (Rendering.JSON.JSON "diffs" SomeJSON)
+parseDiffBuilder DiffSExpression = distributeFoldMap (doDiff (const id) sexprDiff)
+parseDiffBuilder DiffShow        = distributeFoldMap (doDiff (const id) showDiff)
+parseDiffBuilder DiffDotGraph    = distributeFoldMap (doDiff (const id) dotGraphDiff)
 
-jsonDiff :: (DiffEffects sig m) => RenderJSON m syntax -> BlobPair -> m (Rendering.JSON.JSON "diffs" SomeJSON)
-jsonDiff f blobPair = doDiff blobPair (const id) f `catchError` jsonError blobPair
+jsonDiff :: DiffEffects sig m => BlobPair -> m (Rendering.JSON.JSON "diffs" SomeJSON)
+jsonDiff blobPair = doDiff (const id) (pure . jsonTreeDiff blobPair) blobPair `catchError` jsonError blobPair
```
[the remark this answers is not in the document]
robrix: I’m working on this in a follow-up PR, I’ll see what I can do there.

src/Semantic/Api/Diffs.hs
```diff
@@ -53,83 +56,139 @@ data DiffOutputFormat
...
-  , Traversable
-  , ToJSONFields1
-  ]
-
-doDiff :: (DiffEffects sig m)
-  => BlobPair -> Decorate Loc ann -> (forall syntax . CanDiff syntax => BlobPair -> Diff syntax ann ann -> m output) -> m output
-doDiff blobPair decorate render = do
+
+class ShowDiff diff where
+  showDiff :: (Carrier sig m, Member (Reader Config) sig) => diff Loc Loc -> m Builder
```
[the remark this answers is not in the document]
robrix: It is intended to use `Show`, so I don’t want to name it anything that obscures that fact.

src/Semantic/Api/Symbols.hs
robrix: Single-instance typeclasses give you warnings without it.

src/Semantic/Api/Terms.hs
robrix: Again, it’s specifically supposed to use `Show`.

PR #304, opened 2019-10-02
List parsers independently

robrix:
This PR defines a consistent, reusable, abstract mechanism for working with parsers producing arbitrary term types via constraints on the terms. Goals:

- **Consistency:** We shouldn’t run the risk of mapping JavaScript to the TSX parser in one place, and to the TypeScript parser in another.
- **Flexibility:** It should be possible to develop new features and quickly spin them up on actual parse trees.
- **Incrementality:** We must be able to develop support for new languages incrementally; we shouldn’t have to wait til we can do _everything_ before we can do _anything_. This is especially relevant as we migrate from à la carte to precise ASTs, as it must further accommodate multiple ways of expressing existing languages’ terms on a feature-by-feature basis.
- **Convenience:** There should be ready-made accommodations for selecting from well-known sets of parsers.
- **Compile times:** We don’t want to pay high compile-time costs for flexibility & incrementality; our compile times are high enough already.

### DIFF
--- a/src/Control/Effect/Parse.hs
+++ b/src/Control/Effect/Parse.hs
@@ -4,13 +4,16 @@ module Control.Effect.Parse
   Parse(..)
 , parse
 , parseWith
+, parsePairWith
 ) where
 
 import Control.Effect.Carrier
 import Control.Effect.Error
 import Control.Exception (SomeException)
+import Data.Bifunctor.Join
 import Data.Blob
 import Data.Language
+import Data.These
 import Parsing.Parser
 
 data Parse m k
@@ -42,3 +45,13 @@ parseWith
 parseWith parsers with blob = case lookup (blobLanguage blob) parsers of
   Just (SomeParser parser) -> parse parser blob >>= with
   _                        -> noLanguageForBlob (blobPath blob)
+
+parsePairWith
+  :: (Carrier sig m, Member (Error SomeException) sig, Member Parse sig)
+  => [(Language, SomeParser c ann)]
+  -> (forall term . c term => Join These (term ann) -> m a)
+  -> BlobPair
+  -> m a
+parsePairWith parsers with blobPair = case lookup (languageForBlobPair blobPair) parsers of
+  Just (SomeParser parser) -> traverse (parse parser) blobPair >>= with
+  _                        -> noLanguageForBlob (pathForBlobPair blobPair)
--- a/src/Control/Effect/Parse.hs
+++ b/src/Control/Effect/Parse.hs
@@ -40,9 +40,9 @@ parse parser blob = send (Parse parser blob pure)
 -- | Parse a 'Blob' with one of the provided parsers, and run an action on the abstracted term.
 parseWith
   :: (Carrier sig m, Member (Error SomeException) sig, Member Parse sig)
-  => Map.Map Language (SomeParser c ann)
-  -> (forall term . c term => term ann -> m a)
-  -> Blob
+  => Map.Map Language (SomeParser c ann)       -- ^ The set of parsers to select from.
+  -> (forall term . c term => term ann -> m a) -- ^ A function to run on the parsed term. Note that the term is abstract, but constrained by @c@, allowing you to do anything @c@ allows, and requiring that all the input parsers produce terms supporting @c@.
+  -> Blob                                      -- ^ The blob to parse.
   -> m a
 parseWith parsers with blob = case Map.lookup (blobLanguage blob) parsers of
   Just (SomeParser parser) -> parse parser blob >>= with
@@ -51,9 +51,9 @@ parseWith parsers with blob = case Map.lookup (blobLanguage blob) parsers of
 -- | Parse a 'BlobPair' with one of the provided parsers, and run an action on the abstracted term pair.
 parsePairWith
   :: (Carrier sig m, Member (Error SomeException) sig, Member Parse sig)
-  => Map.Map Language (SomeParser c ann)
-  -> (forall term . c term => These (term ann) (term ann) -> m a)
-  -> BlobPair
+  => Map.Map Language (SomeParser c ann)                          -- ^ The set of parsers to select from.
+  -> (forall term . c term => These (term ann) (term ann) -> m a) -- ^ A function to run on the parsed terms. Note that the terms are abstract, but constrained by @c@, allowing you to do anything @c@ allows, and requiring that all the input parsers produce terms supporting @c@.
+  -> BlobPair                                                     -- ^ The blob pair to parse.
   -> m a
 parsePairWith parsers with blobPair = case Map.lookup (languageForBlobPair blobPair) parsers of
   Just (SomeParser parser) -> traverse (parse parser) blobPair >>= with . runJoin
--- a/src/Control/Effect/Parse.hs
+++ b/src/Control/Effect/Parse.hs
@@ -13,6 +13,7 @@ import Control.Exception (SomeException)
 import Data.Bifunctor.Join
 import Data.Blob
 import Data.Language
+import qualified Data.Map as Map
 import Data.These
 import Parsing.Parser
 
@@ -38,20 +39,20 @@ parse parser blob = send (Parse parser blob pure)
 
 parseWith
   :: (Carrier sig m, Member (Error SomeException) sig, Member Parse sig)
-  => [(Language, SomeParser c ann)]
+  => Map.Map Language (SomeParser c ann)
   -> (forall term . c term => term ann -> m a)
   -> Blob
   -> m a
-parseWith parsers with blob = case lookup (blobLanguage blob) parsers of
+parseWith parsers with blob = case Map.lookup (blobLanguage blob) parsers of
   Just (SomeParser parser) -> parse parser blob >>= with
   _                        -> noLanguageForBlob (blobPath blob)
 
 parsePairWith
   :: (Carrier sig m, Member (Error SomeException) sig, Member Parse sig)
-  => [(Language, SomeParser c ann)]
+  => Map.Map Language (SomeParser c ann)
   -> (forall term . c term => These (term ann) (term ann) -> m a)
   -> BlobPair
   -> m a
-parsePairWith parsers with blobPair = case lookup (languageForBlobPair blobPair) parsers of
+parsePairWith parsers with blobPair = case Map.lookup (languageForBlobPair blobPair) parsers of
   Just (SomeParser parser) -> traverse (parse parser) blobPair >>= with . runJoin
   _                        -> noLanguageForBlob (pathForBlobPair blobPair)
--- a/src/Control/Effect/Parse.hs
+++ b/src/Control/Effect/Parse.hs
@@ -37,6 +37,7 @@ parse :: (Member Parse sig, Carrier sig m)
 parse parser blob = send (Parse parser blob pure)
 
 
+-- | Parse a 'Blob' with one of the provided parsers, and run an action on the abstracted term.
 parseWith
   :: (Carrier sig m, Member (Error SomeException) sig, Member Parse sig)
   => Map.Map Language (SomeParser c ann)
@@ -47,6 +48,7 @@ parseWith parsers with blob = case Map.lookup (blobLanguage blob) parsers of
   Just (SomeParser parser) -> parse parser blob >>= with
   _                        -> noLanguageForBlob (blobPath blob)
 
+-- | Parse a 'BlobPair' with one of the provided parsers, and run an action on the abstracted term pair.
 parsePairWith
   :: (Carrier sig m, Member (Error SomeException) sig, Member Parse sig)
   => Map.Map Language (SomeParser c ann)
--- a/src/Control/Effect/Parse.hs
+++ b/src/Control/Effect/Parse.hs
@@ -49,9 +49,9 @@ parseWith parsers with blob = case lookup (blobLanguage blob) parsers of
 parsePairWith
   :: (Carrier sig m, Member (Error SomeException) sig, Member Parse sig)
   => [(Language, SomeParser c ann)]
-  -> (forall term . c term => Join These (term ann) -> m a)
+  -> (forall term . c term => These (term ann) (term ann) -> m a)
   -> BlobPair
   -> m a
 parsePairWith parsers with blobPair = case lookup (languageForBlobPair blobPair) parsers of
-  Just (SomeParser parser) -> traverse (parse parser) blobPair >>= with
+  Just (SomeParser parser) -> traverse (parse parser) blobPair >>= with . runJoin
   _                        -> noLanguageForBlob (pathForBlobPair blobPair)
--- a/src/Control/Effect/Parse.hs
+++ b/src/Control/Effect/Parse.hs
@@ -1,12 +1,16 @@
-{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE ConstraintKinds, ExistentialQuantification, GADTs, RankNTypes #-}
 module Control.Effect.Parse
 ( -- * Parse effect
   Parse(..)
 , parse
+, parseWith
 ) where
 
 import Control.Effect.Carrier
+import Control.Effect.Error
+import Control.Exception (SomeException)
 import Data.Blob
+import Data.Language
 import Parsing.Parser
 
 data Parse m k
@@ -27,3 +31,14 @@ parse :: (Member Parse sig, Carrier sig m)
       -> Blob
       -> m term
 parse parser blob = send (Parse parser blob pure)
+
+
+parseWith
+  :: (Carrier sig m, Member (Error SomeException) sig, Member Parse sig)
+  => [(Language, SomeParser c ann)]
+  -> (forall term . c term => term ann -> m a)
+  -> Blob
+  -> m a
+parseWith parsers with blob = case lookup (blobLanguage blob) parsers of
+  Just (SomeParser parser) -> parse parser blob >>= with
+  _                        -> noLanguageForBlob (blobPath blob)
--- a/src/Diffing/Interpreter.hs
+++ b/src/Diffing/Interpreter.hs
@@ -33,7 +33,7 @@ stripDiff = bimap snd snd
 class HasDiffFor (term :: * -> *) where
   type DiffFor term = (res :: * -> * -> *) | res -> term
 
-class HasDiffFor term => DiffTerms term where
+class (Bifoldable (DiffFor term), HasDiffFor term) => DiffTerms term where
   -- | Diff a 'These' of terms.
   diffTermPair :: These (term ann1) (term ann2) -> DiffFor term ann1 ann2
 
--- a/src/Diffing/Interpreter.hs
+++ b/src/Diffing/Interpreter.hs
@@ -30,7 +30,7 @@ stripDiff :: Functor syntax
 stripDiff = bimap snd snd
 
 class (Bifoldable (DiffFor term)) => DiffTerms term where
-  type DiffFor term = (res :: * -> * -> *) | res -> term
+  type DiffFor term = (diff :: * -> * -> *) | diff -> term
 
   -- | Diff a 'These' of terms.
   diffTermPair :: These (term ann1) (term ann2) -> DiffFor term ann1 ann2
--- a/src/Diffing/Interpreter.hs
+++ b/src/Diffing/Interpreter.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving, TypeFamilyDependencies, TypeOperators, UndecidableInstances #-}
 module Diffing.Interpreter
 ( diffTerms
-, HasDiffFor(..)
 , DiffTerms(..)
 , stripDiff
 ) where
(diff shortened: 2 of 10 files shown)


src/Control/Effect/Parse.hs
```diff
@@ -27,3 +34,24 @@ parse :: (Member Parse sig, Carrier sig m)
       -> Blob
       -> m term
 parse parser blob = send (Parse parser blob pure)
+
+
+parseWith
+  :: (Carrier sig m, Member (Error SomeException) sig, Member Parse sig)
```
patrickt: Can we put the `::` on the same line? Doing so enables me to grep for `parseWith ::` and know that I’m going to jump to (probably) the correct definition.

I would also love to have this yield `Error NoLanguageForBlob` rather than `SomeException`, but we don’t necessarily have to fix that part right now.

src/Control/Effect/Parse.hs
```diff
@@ -27,3 +34,24 @@ parse :: (Member Parse sig, Carrier sig m)
...
+  => [(Language, SomeParser c ann)]
+  -> (forall term . c term => term ann -> m a)
```
patrickt: Can you document what this is for (it’s an algebra, correct?)

robrix:
Some deltas on compile times (taken from `-dshow-passes`):
- `Semantic.Api.Diffs`: faster by ~5.5s (mostly codegen)
- `Semantic.Api.Symbols` faster by ~6s (typechecking & codegen)
- `Semantic.Api.Terms` faster by ~3.5s (typechecking & codegen)

src/Control/Effect/Parse.hs
```diff
@@ -27,3 +34,24 @@ parse :: (Member Parse sig, Carrier sig m)
...
+  -> Blob
+  -> m a
+parseWith parsers with blob = case lookup (blobLanguage blob) parsers of
+  Just (SomeParser parser) -> parse parser blob >>= with
+  _                        -> noLanguageForBlob (blobPath blob)
+
+parsePairWith
+  :: (Carrier sig m, Member (Error SomeException) sig, Member Parse sig)
```
patrickt: Ditto here.

src/Diffing/Interpreter.hs
```diff
@@ -29,11 +30,17 @@ stripDiff :: Functor syntax
           -> Diff.Diff syntax ann1 ann2
 stripDiff = bimap snd snd
 
-class DiffTerms term diff | diff -> term, term -> diff where
+class HasDiffFor (term :: * -> *) where
+  type DiffFor term = (res :: * -> * -> *) | res -> term
```
patrickt: `TypeFamilyDependencies`… you truly do not mess around, @robrix ❤️

Can you indicate in a comment why this dependency is necessary? It’s not obvious from the context.

src/Parsing/Parser.hs
```diff
@@ -21,8 +20,25 @@ module Parsing.Parser
...
+, pythonParserALaCarte'
+, pythonParserPrecise'
+, pythonParser'
+, rubyParser'
+, tsxParser'
+, typescriptParser'
+  -- * Canonical sets of parsers
+, aLaCarteParsers
+, preciseParsers
+, allParsers
```
patrickt: An idea: split this module up into `Parsing.ALaCarte`, `Parsing.Precise`, and `Parsing.Abstract`? That way you don’t need all these ticks.

src/Parsing/Parser.hs
```diff
@@ -62,26 +77,26 @@ import           TreeSitter.Unmarshal
 
 
 -- | A parser, suitable for program analysis, for some specific language, producing 'Term's whose syntax satisfies a list of typeclass constraints.
-data SomeAnalysisParser typeclasses ann where
-  SomeAnalysisParser :: ( ApplyAll typeclasses (Sum fs)
+data SomeAnalysisParser constraint ann where
```
patrickt: Explicit `:: Constraint` kind signature on `constraint` here, just for safety’s sake?

src/Parsing/Parser.hs
```diff
@@ -62,26 +77,26 @@ import           TreeSitter.Unmarshal
...
+                      , constraint (Sum PHP.Syntax)
+                      , constraint (Sum Python.Syntax)
+                      , constraint (Sum Ruby.Syntax)
+                      , constraint (Sum TypeScript.Syntax)
+                      , constraint (Sum Haskell.Syntax)
                       )
-                   => proxy typeclasses                  -- ^ A proxy for the list of typeclasses required, e.g. @(Proxy :: Proxy '[Show1])@.
-                   -> Language                           -- ^ The 'Language' to select.
-                   -> SomeAnalysisParser typeclasses Loc -- ^ A 'SomeAnalysisParser' abstracting the syntax type to be produced.
+                   => proxy constraint                  -- ^ A proxy for the list of typeclasses required, e.g. @(Proxy :: Proxy '[Show1])@.
```
patrickt: This documentation is out of date.

src/Semantic/Api/Diffs.hs
```diff
@@ -157,51 +176,29 @@ instance (Foldable syntax, Functor syntax, HasDeclaration syntax) => SummarizeDi
...
+diffWith parsers render blobPair = parsePairWith parsers (render <=< diffTerms blobPair) blobPair
+
+decoratingDiffWith
+  :: forall ann c output m sig
+  .  (forall term . c term => DiffTerms term, DiffEffects sig m)
+  => [(Language, SomeParser c Loc)]
+  -> (forall term . c term => Blob -> term Loc -> term ann)
+  -> (forall term . c term => DiffFor term ann ann -> m output)
+  -> BlobPair
+  -> m output
```
patrickt: Can you add some per-parameter Haddocks to clarify what each of these does?

src/Semantic/Api/Terms.hs
```diff
@@ -85,15 +85,21 @@ quietTerm blob = showTiming blob <$> time' ( doParse (fmap (const (Right ())) .
...
+
 class ShowTerm term where
   showTerm :: (Carrier sig m, Member (Reader Config) sig) => term Loc -> m Builder
 
 instance (Functor syntax, Show1 syntax) => ShowTerm (Term syntax) where
   showTerm = serialize Show . quieterm
 
 instance ShowTerm Py.Term where
-  showTerm = serialize Show . Py.getTerm
+  showTerm = serialize Show . (() <$) . Py.getTerm
```
patrickt:
```suggestion
  showTerm = serialize Show . void . Py.getTerm
```

src/Semantic/Graph.hs
```diff
@@ -71,7 +71,8 @@ import           Text.Show.Pretty (ppShow)
 
 data GraphType = ImportGraph | CallGraph
 
-type AnalysisClasses = '[ Declarations1, Eq1, Evaluatable, FreeVariables1, AccessControls1, Foldable, Functor, Ord1, Show1 ]
+class (Declarations1 syntax, Eq1 syntax, Evaluatable syntax, FreeVariables1 syntax, AccessControls1 syntax, Foldable syntax, Functor syntax, Ord1 syntax, Show1 syntax) => AnalysisClasses syntax
+instance (Declarations1 syntax, Eq1 syntax, Evaluatable syntax, FreeVariables1 syntax, AccessControls1 syntax, Foldable syntax, Functor syntax, Ord1 syntax, Show1 syntax) => AnalysisClasses syntax
```
patrickt: Newlines here, please: lotta parameters!

src/Tags/Tagging.hs
```diff
@@ -23,11 +23,11 @@ import           Tags.Taggable
 
 runTagging :: (IsTaggable syntax)
            => Language
-           -> Source.Source
            -> [Text]
+           -> Source.Source
```
patrickt: This parameter order change gives it better compositionality, I’m assuming?

src/Semantic/Api/Diffs.hs
```diff
@@ -157,51 +176,29 @@ instance (Foldable syntax, Functor syntax, HasDeclaration syntax) => SummarizeDi
...
+decoratingDiffWith parsers decorate render blobPair = parsePairWith parsers (render <=< diffTerms blobPair . bimap (decorate blobL) (decorate blobR)) blobPair where
+  (blobL, blobR) = fromThese errorBlob errorBlob (runJoin blobPair)
+  errorBlob = Prelude.error "evaluating blob on absent side"
```
patrickt: Is it possible to throw either a fused-effects `Error` or, if this error is not recoverable, create an exception type and `throw` that?

src/Control/Effect/Parse.hs
robrix: No, it’s just a function you run on an abstracted term under the constraint `c`.

src/Control/Effect/Parse.hs
robrix: This is intended to yield stable diffs under renames. I’d really rather we used this style given how much re-indentation we tend to waste time doing and reading.

src/Parsing/Parser.hs
robrix: Well, we’d end up having to import more things in a bunch of places (`Parsing.Parser` couldn’t import those since they’d need `SomeParser` & `Parser`), and at the end of the day, the goal is for all the unticked versions to go away. I’d rather leave the ticks as an incentive to remove the à la carte parsers altogether 😊

src/Parsing/Parser.hs
robrix: Good catch.

src/Tags/Tagging.hs
robrix: Yep.

src/Semantic/Api/Diffs.hs
robrix: It’s impossible for this to happen; the shape of the `These` in the `BlobPair` determines the shape of the `These` that this is eliminating. Thus, we’re guaranteed to have a blob on a side if we have a term on that side, so this is actually total.

src/Parsing/Parser.hs
```diff
@@ -112,12 +131,6 @@ data Parser term where
   MarkdownParser :: Parser (Term (TermF [] CMarkGFM.NodeType) (Node Markdown.Grammar))
 
 
--- | Apply all of a list of typeclasses to all of a list of functors using 'Apply'. Used by 'someParser' to constrain all of the language-specific syntax types to the typeclasses in question.
-type family ApplyAll (typeclasses :: [(* -> *) -> Constraint]) (syntax :: * -> *) :: Constraint where
-  ApplyAll (typeclass ': typeclasses) syntax = (typeclass syntax, ApplyAll typeclasses syntax)
-  ApplyAll '[] syntax = ()
```
robrix: Good riddance!

src/Parsing/Parser.hs
```diff
@@ -189,3 +202,139 @@ someASTParser PHP        = Just (SomeASTParser (ASTParser tree_sitter_php :: Par
...
 someASTParser Markdown   = Nothing
 someASTParser Unknown    = Nothing
+
+
+-- $abstract
+-- Most of our features are intended to operate over multiple languages, each represented by disjoint term types. Thus, we typically implement them using typeclasses, allowing us to share a single interface to invoke the feature, while specializing the implementation(s) as appropriate for each distinct term type.
+--
+-- In order to accomplish this, we employ 'SomeParser', which abstracts over parsers of various term types, while ensuring that some desired constraint holds. Constructing a @'SomeParser' c@ requires satisfiyng the constraint @c@ against the underlying 'Parser'’s term type, and so it can be used to parse with any of a map of parsers whose terms support @c@.
+--
+-- In practice, this means using 'Control.Effect.Parse.parseWith', and passing in a map of parsers to select from for your feature. It is recommended to define the map as a concrete top-level binding using the abstract parsers or ideally the canonical maps of parsers, below; using the abstracted parsers or canonical maps directly with 'Control.Effect.Parse.parseWith' will lead to significantly slower compiles.
```
robrix: This is probably the most important part of this PR, as it gives an overview of the approach to supporting multiple languages and features flexibly, while allowing us to adopt support for them incrementally, and while keeping compiles tractable.

src/Parsing/Parser.hs
```diff
@@ -189,3 +202,139 @@ someASTParser PHP        = Just (SomeASTParser (ASTParser tree_sitter_php :: Par
...
+-- @
+--
+-- Good:
+--
+-- @
+-- fancyParsers :: 'Map' 'Language' ('SomeParser' Fancy 'Loc')
+-- fancyParsers = preciseParsers
+--
+-- isFancy :: (Carrier sig m, Member Parse sig) => Blob -> m Bool
+-- isFancy = parseWith fancyParsers (pure . isTermFancy) -- much faster compiles
```
robrix: Fancy terms are like fancy types, but more plebeian.

src/Parsing/Parser.hs
```diff
@@ -189,3 +202,139 @@ someASTParser PHP        = Just (SomeASTParser (ASTParser tree_sitter_php :: Par
...
+pythonParserALaCarte' :: c (Term (Sum Python.Syntax)) => (Language, SomeParser c Loc)
+pythonParserALaCarte' = (Python, SomeParser pythonParser)
+
+pythonParserPrecise' :: c Py.Term => (Language, SomeParser c Loc)
+pythonParserPrecise' = (Python, SomeParser pythonParserPrecise)
+
+pythonParser' :: (c (Term (Sum Python.Syntax)), c Py.Term) => PerLanguageModes -> (Language, SomeParser c Loc)
+pythonParser' modes = case pythonMode modes of
+  ALaCarte -> (Python, SomeParser pythonParser)
+  Precise  -> (Python, SomeParser pythonParserPrecise)
```
robrix: When we’re finished migrating support for Python over to the precise stuff, then this will go away and we’ll only have the precise version ✨

src/Semantic/Api/Diffs.hs
```diff
@@ -58,12 +58,12 @@ data DiffOutputFormat
 parseDiffBuilder :: (Traversable t, DiffEffects sig m) => DiffOutputFormat -> t BlobPair -> m Builder
 parseDiffBuilder DiffJSONTree    = distributeFoldMap jsonDiff >=> serialize Format.JSON -- NB: Serialize happens at the top level for these two JSON formats to collect results of multiple blob pairs.
 parseDiffBuilder DiffJSONGraph   = diffGraph >=> serialize Format.JSON
-parseDiffBuilder DiffSExpression = distributeFoldMap (doDiff (const id) sexprDiff)
-parseDiffBuilder DiffShow        = distributeFoldMap (doDiff (const id) showDiff)
-parseDiffBuilder DiffDotGraph    = distributeFoldMap (doDiff (const id) dotGraphDiff)
+parseDiffBuilder DiffSExpression = distributeFoldMap (diffWith sexprDiffParsers sexprDiff)
+parseDiffBuilder DiffShow        = distributeFoldMap (diffWith showDiffParsers showDiff)
+parseDiffBuilder DiffDotGraph    = distributeFoldMap (diffWith dotGraphDiffParsers dotGraphDiff)
```
robrix: Pretty happy with how this turned out. Great suggestion to obviate the need for the `(const id)`, @patrickt!

src/Semantic/Api/Diffs.hs
```diff
@@ -81,20 +81,24 @@ diffGraph blobs = DiffTreeGraphResponse . V.fromList . toList <$> distributeFor
...
 
-type Decorate a b = forall term diff . DiffActions term diff => Blob -> term a -> term b
 
+dotGraphDiffParsers :: Map Language (SomeParser DOTGraphDiff Loc)
+dotGraphDiffParsers = aLaCarteParsers
 
-class DOTGraphDiff diff where
-  dotGraphDiff :: (Carrier sig m, Member (Reader Config) sig) => diff Loc Loc -> m Builder
+class DiffTerms term => DOTGraphDiff term where
+  dotGraphDiff :: (Carrier sig m, Member (Reader Config) sig) => DiffFor term Loc Loc -> m Builder
```
robrix: The new abstract parsing facilities require the constraints be phrased against the term, so we rephrase all of these classes against the term type, using `DiffTerms` as a supertype, and a type family dependency on `DiffFor` in order to ensure that `dotGraphDiff` is unambiguous despite only using `term` as a parameter to a type family.

I feel a bit like I’ve gained a level, in terms of using `-XTypeFamilyDependencies` successfully for the first time, but regardless of that, I’m also just quite happy with how this solution has turned out, since it means we don’t have to have separate abstraction facilities for diffs and terms any more.

The essential realization was the term/term pairs was the wrong place to place the abstraction; placing it on the _parser_ enables us to generate terms/term pairs from the same place with the assurance that a) the term type is the same between both, and b) the constraint `c` holds for both. And since `c term` in turn allows us to constrain the use of `DiffFor term`, all our term/diff processing functionality can share a small set of primitives for abstract parsing.

src/Semantic/Api/Diffs.hs
```diff
@@ -81,20 +81,24 @@ diffGraph blobs = DiffTreeGraphResponse . V.fromList . toList <$> distributeFor
...
 
-instance (ConstructorName syntax, Foldable syntax, Functor syntax) => DOTGraphDiff (Diff syntax) where
+instance (ConstructorName syntax, Diffable syntax, Eq1 syntax, Hashable1 syntax, Traversable syntax) => DOTGraphDiff (Term syntax) where
   dotGraphDiff = serialize (DOT (diffStyle "diffs")) . renderTreeGraph
 
 
-class JSONGraphDiff diff where
-  jsonGraphDiff :: BlobPair -> diff Loc Loc -> DiffTreeFileGraph
+jsonGraphDiffParsers :: Map Language (SomeParser JSONGraphDiff Loc)
+jsonGraphDiffParsers = aLaCarteParsers
```
robrix: As documented in `Parsing.Parser`, it’s quite important that this be of concrete type, and thus a CAF: don’t return it in an abstract `Monad`, don’t generalize it over the constraint type. Otherwise your compile times go through the roof, as evinced by `-dshow-passes`.

src/Semantic/Api/Symbols.hs
```diff
@@ -90,38 +92,15 @@ tagToSymbol Tag{..} = Symbol
...
-      | otherwise                   -> parse Parser.pythonParser        blob >>= with . mkTerm
-    Ruby       -> parse Parser.rubyParser       blob >>= with . mkTerm
-    TypeScript -> parse Parser.typescriptParser blob >>= with . mkTerm
-    TSX        -> parse Parser.tsxParser        blob >>= with . mkTerm
-    PHP        -> parse Parser.phpParser        blob >>= with . mkTerm
-    _          -> noLanguageForBlob (blobPath blob)
-    where mkTerm :: Term syntax Loc -> ALaCarteTerm syntax Loc
-          mkTerm = ALaCarteTerm (blobLanguage blob) symbolsToSummarize
+class ToTags t where
+  tags :: Language -> [Text] -> Source -> t Loc -> [Tag]
```
robrix: `ALaCarteTerm` prevented us being able to use the new parser abstraction facilities, so I’ve instead _further_ abstracted tag generation behind this local class, implemented using `runTagging` on the à la carte side and `Precise.ToTags` on the precise side.

src/Semantic/Api/Terms.hs
```diff
@@ -126,28 +141,3 @@ instance (Foldable syntax, Functor syntax, ConstructorName syntax) => JSONGraphT
...
-  JavaScript -> parse tsxParser        blob >>= with
-  JSON       -> parse jsonParser       blob >>= with
-  JSX        -> parse tsxParser        blob >>= with
-  Markdown   -> parse markdownParser   blob >>= with
-  Python     -> parse pythonParser     blob >>= with
-  Ruby       -> parse rubyParser       blob >>= with
-  TypeScript -> parse typescriptParser blob >>= with
-  TSX        -> parse tsxParser        blob >>= with
-  PHP        -> parse phpParser        blob >>= with
-  _          -> noLanguageForBlob (blobPath blob)
```
robrix: 🔥

src/Semantic/Graph.hs
```diff
@@ -71,7 +71,29 @@ import           Text.Show.Pretty (ppShow)
...
+  ( Declarations1 syntax
+  , Eq1 syntax
+  , Evaluatable syntax
+  , FreeVariables1 syntax
+  , AccessControls1 syntax
+  , Foldable syntax
+  , Functor syntax
+  , Ord1 syntax
+  , Show1 syntax
+  ) => AnalysisClasses syntax
```
robrix: I’m not a huge fan of this, but it allows us to cut the number of generated coercions by 4/5ths. On the other hand, that doesn’t seem to do a lot to compile times, at least at `-O1`.

We can’t use the abstracted parsing facilities for this module, because far too much of the machinery is coupled to `Term` to be able to usefully abstract it. `Evaluatable` itself takes a constructor parameterized by subterms, which won’t work for precise ASTs. But that’s alright, because we wouldn’t want to anyway—the whole idea is for precise ASTs to be compiled to core and analyzed there instead.

(The same goes for the `Semantic.AST` stuff, which I’d honestly rather we just got rid of outright.)

src/Parsing/Parser.hs
```diff
@@ -189,3 +202,139 @@ someASTParser PHP        = Just (SomeASTParser (ASTParser tree_sitter_php :: Par
...
+     , c (Term (Sum Haskell.Syntax))
+     , c (Term (Sum JSON.Syntax))
+     , c (Term (Sum Markdown.Syntax))
+     , c (Term (Sum PHP.Syntax))
+     , c (Term (Sum Python.Syntax))
+     , c (Term (Sum Ruby.Syntax))
+     , c (Term (Sum TSX.Syntax))
+     , c (Term (Sum TypeScript.Syntax))
+     )
+  => Map Language (SomeParser c Loc)
```
patrickt: Is it worth creating a newtype/synonym for a map from `Language` to `SomeParser` values?

src/Parsing/Parser.hs
```diff
@@ -189,3 +202,139 @@ someASTParser PHP        = Just (SomeASTParser (ASTParser tree_sitter_php :: Par
...
+  , phpParser'
+  , pythonParserALaCarte'
+  , rubyParser'
+  , typescriptParser'
+  , tsxParser'
+  ]
+
+-- | The canonical set of parsers producing precise terms.
+preciseParsers :: c Py.Term => Map Language (SomeParser c Loc)
+preciseParsers = Map.fromList
```
patrickt: Nit: you could turn on `OverloadedLists`, if you wanted.

src/Semantic/Api/Diffs.hs
patrickt: tfw Rob gains a level

src/Semantic/Api/Diffs.hs
patrickt: Okay, cool, `error` is the right thing then.

src/Parsing/Parser.hs
robrix: Dunno. I think we’ve got as many of them as we’re likely to any time soon, at least.

src/Parsing/Parser.hs
robrix: I do _not_ want 😆

PR #310, opened 2019-10-04
Remove Haskell language modules for now

rewinfrey:
Relates to https://github.com/tree-sitter/tree-sitter-haskell/issues/27. These modules are not able to be used due to the current state of `tree-sitter-haskell`, and only add to the compile / build times of `semantic`. Removing these for now and plan to revisit in the future.


README.md
```diff
@@ -86,7 +86,7 @@ Available options:
 |          | Java           | ✅     | ✅     | ✅    | 🔶  | ✅      |               | | |
 |          | JSON           | ✅     | ✅     | ✅    | N/A | N/A     | N/A          | N/A| |
 |          | JSX            | ✅     | ✅     | ✅    | 🔶  |         |              | | |
-|          | Haskell        | ✅     | ✅     | ✅    | 🔶  |  ✅       |              | | |
+|          | Haskell        | 🚧     | 🚧     | 🚧    | 🔶  | 🚧       |              | | |
```
robrix: tyvm!

PR #317, opened 2019-10-07
Use semantic-source in semantic-core & represent Name as a newtype

robrix:
This PR replaces the `Span` type in `semantic-core` with the one from `semantic-source`, and represents `Name` as a `newtype` instead of just a type synonym.


semantic-core/semantic-core.cabal
```diff
@@ -25,6 +25,7 @@ library
                      , Analysis.ImportGraph
                      , Analysis.ScopeGraph
                      , Analysis.Typecheck
+                     , Control.Carrier.Fail.WithLoc
```
robrix: I also moved `FailWithLocC` into its own module.

semantic-core/src/Control/Carrier/Fail/WithLoc.hs
```diff
@@ -0,0 +1,31 @@
...
+  deriving (Alternative, Applicative, Functor, Monad)
+
+instance (Carrier sig m, Effect sig, Member (Reader Loc) sig) => MonadFail (FailC m) where
+  fail s = do
+    loc <- ask
+    FailC (throwError (loc :: Loc, s))
+
+instance (Carrier sig m, Effect sig, Member (Reader Loc) sig) => Carrier (Fail :+: sig) (FailC m) where
+  eff (L (Fail s)) = fail s
+  eff (R other)    = FailC (eff (R (handleCoercible other)))
```
robrix: This definition is unchanged, it’s just been pulled here because it never really belonged in `Data.Loc`.

semantic-core/src/Data/Loc.hs
```diff
@@ -28,28 +17,7 @@ data Loc = Loc
   deriving (Eq, Ord, Show)
 
 interactive :: Loc
-interactive = Loc "<interactive>" emptySpan
-
-data Span = Span
-  { spanStart :: {-# UNPACK #-} !Pos
-  , spanEnd   :: {-# UNPACK #-} !Pos
-  }
-  deriving (Eq, Ord, Show)
```
robrix: We now depend on `semantic-source` and as such get `Span` & `Pos` from there.

semantic-core/src/Data/Loc.hs
```diff
@@ -28,28 +17,7 @@ data Loc = Loc
...
-
-instance Pretty Span where
-  pretty (Span s e) = pretty s <> "-" <> pretty e
```
robrix: We never used this and we never should.

semantic-core/src/Data/Name.hs
```diff
@@ -15,10 +15,14 @@ module Data.Name
...
 import           Data.HashSet (HashSet)
 import qualified Data.HashSet as HashSet
+import           Data.String
 import           Data.Text as Text (Text, any, unpack)
+import           Data.Text.Prettyprint.Doc
 
 -- | User-specified and -relevant names.
-type Name = Text
+newtype Name = Name { unName :: Text }
+  deriving (Eq, IsString, Ord, Pretty, Show)
```
robrix: At @patrickt’s behest, this is now a `newtype`; should improve error messages.

semantic-python/src/Language/Python/Core.hs
```diff
@@ -97,9 +96,8 @@ compile :: ( Compile py
         -> m (t Name)
 compile t = compileCC t (pure none)
 
-locFromTSSpan :: SourcePath -> Source.Span -> Loc
-locFromTSSpan fp (Source.Span (Source.Pos a b) (Source.Pos c d))
-  = Data.Loc.Loc (rawPath fp) (Data.Loc.Span (Data.Loc.Pos a b) (Data.Loc.Pos c d))
+locFromTSSpan :: SourcePath -> Span -> Loc
+locFromTSSpan fp = Data.Loc.Loc (rawPath fp)
```
robrix: I want to change how annotations work so as to not require this sort of thing but for now this will do.

robrix:
I totally clobbered #316 😭 whoops

PR #334, opened 2019-10-11
Rename Core.File to Analysis.File

robrix:
This PR:

- [x] Renames `Core.File` to `Analysis.File`, because it relates to analysis rather than to the core language.
- [x] Depends on #327.
- [x] Depends on #329.
- [x] Depends on #332.

### DIFF
--- a/semantic-core/semantic-core.cabal
+++ b/semantic-core/semantic-core.cabal
@@ -42,6 +42,7 @@ library
   exposed-modules:
     Analysis.Analysis
     Analysis.Concrete
+    Analysis.File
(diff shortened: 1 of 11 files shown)


semantic-core/semantic-core.cabal
```diff
@@ -53,7 +54,6 @@ library
     Core.Core.Parser
     Core.Core.Pretty
     Core.Eval
-    Core.File
```
robrix: This PR is super straightforward, I just renamed the one file because it relates to the analysis of projects more so than to `Core` itself.

PR #335, opened 2019-10-11
Semantic analysis

robrix:
This PR extracts `semantic-core`’s `Analysis.*` and `Control.*` hierarchies into a new `semantic-analysis` package.

- [x] Depends on #327.
- [x] Depends on #329.
- [x] Depends on #332.
- [x] Depends on #334.


.travis.yml
```diff
@@ -37,7 +37,6 @@ script:
 - cabal new-build
 - cabal new-run semantic:test
 - cabal new-run semantic-core:test
-- cabal new-run semantic-core:doctest
```
robrix: 1. With `Analysis.Concrete` moved into `semantic-analysis`, we no longer have any doctests in `semantic-core`.

2. The doctest in `Analysis.Concrete` involved `semantic-core`’s `eval` & `Core`, and didn’t really belong there; it was quite low value so I’ve 🔥d it for the time being, and will replace it with tests for `semantic-analysis` in future PR(s).

cabal.project
```diff
@@ -11,6 +12,9 @@ jobs: $ncpus
 package semantic
   ghc-options: -Werror
 
+package semantic-analysis
+  ghc-options: -Werror
+
```
robrix: beeeees!

semantic-analysis/src/Analysis/Concrete.hs
```diff
@@ -65,10 +65,6 @@ data Edge = Lexical | Import
   deriving (Eq, Ord, Show)
 
 
--- | Concrete evaluation of a term to a value.
---
---   >>> map fileBody (snd (concrete eval [File (Path.absRel "bool") (Span (Pos 1 1) (Pos 1 5)) (Core.bool True)]))
---   [Right (Bool True)]
```
robrix: As described above, this was a very low-value test, and retaining it would have complicated things significantly, so 🔥

semantic-core/semantic-core.cabal
```diff
@@ -40,53 +40,25 @@ library
...
-    Analysis.ScopeGraph
-    Analysis.Typecheck
-    Control.Carrier.Fail.WithLoc
-    Control.Carrier.Readline.Haskeline
-    Control.Effect.Readline
     Core.Core
-    Core.Core.Parser
-    Core.Core.Pretty
+    Core.Parser
+    Core.Pretty
```
robrix: I took the opportunity to rename these since they’re unambiguous now by virtue of the module hierarchy change in #327.

semantic-core/semantic-core.cabal
```diff
@@ -40,53 +40,25 @@ library
...
     , parsers                     ^>= 0.12.10
     , pathtype                    ^>= 0.8.1
     , prettyprinter               ^>= 1.2.1
     , prettyprinter-ansi-terminal ^>= 1.1.1
+    , semantic-analysis           ^>= 0
     , semantic-source             ^>= 0
-    , semigroupoids               ^>= 5.3
-    , terminal-size               ^>= 0.3
     , text                        ^>= 1.2.3.1
-    , transformers                ^>= 0.5.6
```
robrix: `semigroupoids` and `transformers` were just plain unused ✨

cabal.project
```diff
@@ -1,4 +1,5 @@
 packages: .
+          semantic-analysis
```
patrickt: We’re starting to get a lot of packages in this hierarchy. I don’t think this is a bad thing—indeed, I think this will help us with compile times and tooling reliability—but we should start documenting in the README our notions of how the project is organized, and which responsibilities each subproject covers.

robrix: Yeah, @dcreager and I have talked about pulling (some of) the architecture diagram I drew up into this repo.

Was also thinking we might want to have a languages subdir. Not sure yet. Thoughts?

PR #336, opened 2019-10-11
Readline path types

robrix:
This PR replaces the use of the `directory` & `filepath` packages in `semantic-analysis` with the type-safe `pathtype` package.

- [x] Depends on #332.
- [x] Depends on #334.
- [x] Depends on #335.

### DIFF
--- a/semantic-analysis/src/Control/Carrier/Readline/Haskeline.hs
+++ b/semantic-analysis/src/Control/Carrier/Readline/Haskeline.hs
@@ -23,8 +23,9 @@ import Data.Text.Prettyprint.Doc
 import Data.Text.Prettyprint.Doc.Render.Terminal (renderIO)
 import System.Console.Haskeline hiding (Handler, handle)
 import System.Console.Terminal.Size as Size
-import System.Directory
-import System.FilePath
+import System.Path ((</>))
+import qualified System.Path as Path
+import System.Path.Directory
 import System.IO (stdout)
 
 runReadline :: MonadException m => Prefs -> Settings m -> ReadlineC m a -> m a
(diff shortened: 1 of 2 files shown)


semantic-analysis/src/Control/Carrier/Readline/Haskeline.hs
```diff
@@ -33,11 +34,11 @@ runReadline prefs settings = runInputTWithPrefs prefs (coerce settings) . runM .
...
 runReadlineWithHistory block = do
   homeDir <- liftIO getHomeDirectory
-  prefs <- liftIO $ readPrefs (homeDir </> ".haskeline")
-  let settingsDir = homeDir </> ".local/semantic-core"
+  prefs <- liftIO $ readPrefs (Path.toString (homeDir </> Path.relFile ".haskeline"))
+  let settingsDir = homeDir </> Path.relDir ".local" </> Path.relDir "semantic-core"
       settings = Settings
         { complete = noCompletion
-        , historyFile = Just (settingsDir <> "/repl_history")
+        , historyFile = Just (Path.toString (settingsDir </> Path.relFile "repl_history"))
```
robrix: As @patrickt will tell you, `pathtype`’s docs are really horribly organized, but the essence here is that I’m using constructors specific to a) whether the component I’m adding is relative or absolute, and b) whether it’s expected to be a directory or a file. Then I produce the actual `FilePath` (a `Prelude`-provided synonym for `String`) with `Path.toString`, since `haskeline` (and indeed the OS) doesn’t know about `pathtype`.

maxbrunsfeld:
Just curious, what does the term 'readline' refer to here?

robrix:
> Just curious, what does the term 'readline' refer to here?

We’ve defined a `Readline` effect (basically just exposing reading from the input and printing output, at present), and this carrier for it uses the [`haskeline` package](http://hackage.haskell.org/package/haskeline) (a pure Haskell `readline`-alike package used in `ghci`, among other places) to implement it with full support for history, completion, &c.

We intend to have a REPL for core & analysis at some point, and this is a key part of that. (I’m using essentially the same definition in `path` for exactly that purpose.)

PR #352, opened 2019-10-21
Build & test against 8.6.5 in CI & expire old compilers’ caches

robrix:
This PR:

- [x] Bumps CI to use 8.6.5.
- [x] Updates the `tested-with` fields accordingly where needed.
- [x] 🔥s caches from older compilers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,8 @@ before_cache:
 
 matrix:
   include:
-    - compiler: "ghc-8.6.4"
-      addons: {apt: {packages: [cabal-install-2.4,ghc-8.6.4], sources: [hvr-ghc]}}
+    - compiler: "ghc-8.6.5"
+      addons: {apt: {packages: [cabal-install-2.4,ghc-8.6.5], sources: [hvr-ghc]}}
 
 before_install:
 - mkdir -p $HOME/.local/bin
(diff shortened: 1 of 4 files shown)


semantic-ast/semantic-ast.cabal
```diff
@@ -16,6 +16,8 @@ copyright:           (c) 2019 GitHub, Inc.
 category:            Language
 extra-source-files:  CHANGELOG.md
 
+tested-with:         GHC == 8.6.5
+
```
robrix: cc @aymannadeem

.travis.yml
```diff
@@ -17,10 +17,14 @@ before_cache:
 
   - rm -rfv $HOME/.cabal/packages/head.hackage
 
+  # remove files from old compilers
+  - rm -rfv $HOME/.cabal/store/{!(ghc-8.6.5)}
+  - rm -rfv $TRAVIS_BUILD_DIR/dist-newstyle/build/*/{!(ghc-8.6.5)}
+
```
robrix: We’re taking several minutes just to download and unpack the cache, meaning it’s probably bigger than we might expect. This just clears out the non-8.6.5 contents of those dirs.

robrix:
Built successfully in ~33min; restarted the build to see how it goes.

robrix:
Aaaaand successful in 3.5min!!

robrix:
in particular, ~17s for futzing with the cache, vs. ~393s!

robrix:
~40s for packing & uploading the new cache, vs. ~387s.

I’mma call this a win.

PR #374, opened 2019-11-01
Multi component ghcide setup

robrix:
This PR:

- [x] Depends on #373.
- [x] Adds a couple of scripts for computing the correct flags for `ghcide` to use to load all of our components.
- [x] Adds a `hie.yaml` file configuring `ghcide` thus.


script/ghci-flags
```diff
@@ -0,0 +1,63 @@
...
+echo "-i$root/app"
+echo "-i$root/src"
+echo "-i$root/bench"
+echo "-i$root/test"
+
+echo "-optP-include"
+echo "-optP$root/$repl_builddir/build/x86_64-osx/ghc-$ghc_version/semantic-0.8.0.0/build/autogen/cabal_macros.h"
+
+# Emit package flags from the environment file, removing comments & prefixing with -
+cabal exec --builddir=$repl_builddir -v0 bash -- -c 'cat $GHC_ENVIRONMENT' | grep -v '^--' | sed -e 's/^/-/'
```
robrix: Rather than using the existing `.ghc.environment*` file in the project root, I’ve opted to use `cabal exec` to get the package flags out. I can’t just use `echo "-package-conf=$(cabal exec -v0 bash -- -c 'echo $GHC_ENVIRONMENT')"` because `cabal` makes a new file in `dist-newstyle/tmp` and deletes it as soon as the exec’d program exits, meaning that it would be removed before `ghcide` could find it.

script/ghci-flags
```diff
@@ -0,0 +1,63 @@
...
+echo "-i$root/semantic-ast/src"
+echo "-i$root/semantic-core/src"
+echo "-i$root/semantic-java/src"
+echo "-i$root/semantic-json/src"
+echo "-i$root/semantic-python/src"
+echo "-i$root/semantic-tags/src"
+echo "-i$root/app"
+echo "-i$root/src"
+echo "-i$root/bench"
+echo "-i$root/test"
```
robrix: This is pretty similar to what we were doing in `.ghci.semantic` (added in #373), but distinct in that VS Code runs `ghcide` with the `PWD` env var set to `/` and apparently the `-i` flags are resolved relative to that environment variable rather than the actual, you know, working directory, and thus we have to use absolute paths.

script/ghci-flags
```diff
@@ -0,0 +1,63 @@
...
+  repl_builddir=dist-newstyle
+fi
+
+echo "-O0"
+echo "-ignore-dot-ghci"
+
+echo "-outputdir $repl_builddir/build/x86_64-osx/ghc-$ghc_version/semantic-0.8.0.0/build"
+echo "-odir $repl_builddir/build/x86_64-osx/ghc-$ghc_version/semantic-0.8.0.0/build"
+echo "-hidir $repl_builddir/build/x86_64-osx/ghc-$ghc_version/semantic-0.8.0.0/build"
+echo "-stubdir $repl_builddir/build/x86_64-osx/ghc-$ghc_version/semantic-0.8.0.0/build"
```
robrix: We don’t instruct `ghcide` to use `-fwrite-interface` or `-fobject-code`, but putting these here allows us to share more of the logic for determining the compiler version &c.

script/ghci-flags
```diff
@@ -0,0 +1,63 @@
...
+
+root="$(pwd)"
+ghc_version="$(ghc --numeric-version)"
+
+if [ "$1" == "--builddir" ]; then
+  repl_builddir="$2"
+  shift 2
+else
+  repl_builddir=dist-newstyle
+fi
```
robrix: We can’t pass flags into this program from `hie.yaml`, but I wanted to reuse it from `script/repl` without duplicating much of it or adding redundant wrapper scripts, so I’ve made this default to `dist-newstyle` if it’s unspecified.

script/ghci-flags-dependencies
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+# Computes the paths to files causing changes to the ghci flags. You probably won’t be running this yourself, but rather ghcide will via configuration in hie.yaml.
```
robrix: Changes to the listed files might result in different package flags for `ghcide` to learn of, so we tell it to watch them.

script/repl
```diff
@@ -11,5 +11,5 @@ repl_builddir=dist-repl
 if [[ ! -d $repl_builddir ]]; then
   echo "$repl_builddir does not exist, first run 'cabal repl --builddir=$repl_builddir', exit, and then re-run $0"
 else
-  cabal exec --builddir=$repl_builddir ghci -- -ghci-script=.ghci.semantic $@
+  cabal exec --builddir=$repl_builddir ghci -- -ghci-script=.ghci.semantic $(script/ghci-flags --builddir "$repl_builddir") -no-ignore-dot-ghci $@
```
robrix: 1. We reuse `script/ghci-flags` from `script/repl` in order to share a bunch of the path computation logic.

2. `script/ghci-flags` tells `ghcide` to ignore `.ghci` files so as to avoid getting confused, so we also say `-no-ignore-dot-ghci` afterwards because `script/repl` _should_ load them.

.ghci.semantic
```diff
@@ -1,61 +1,17 @@
 -- GHCI settings for semantic, collected by running cabal repl -v and checking out the flags cabal passes to ghc.
```
robrix: All of the deletions in this file are now being handled via `script/ghci-flags` instead.

hie.yaml
```diff
@@ -0,0 +1,4 @@
+cradle:
+  bios:
+    program: script/ghci-flags
+    dependency-program: script/ghci-flags-dependencies
```
robrix: 1. `hie-bios` recently added support for multiple cabal component cradles, but I couldn’t get it working with `ghcide` yet.

2. My `direct` configuration wasn’t portable due to VS Code using a `PWD` env var of `/` and `-i` flags being resolved relative to that env var rather than the actual working directory. So we had to compute it.

3. Computing it is probably better for our purposes anyway, since it allows us to share work between this and `script/repl`.

script/ghci-flags
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+# Computes the flags for ghcide to pass to ghci. You probably won’t be running this yourself, but rather ghcide will via configuration in hie.yaml.
```
robrix: The approach here is similar to the approach of #373: basically just listing all the options we can think of, collected by clever use of `cabal repl -v`.

A more robust approach would probably be to do something like `cabal repl` for each component and union together the options, but that actually seems pretty challenging:

- stuff like package flags come in as two flags
- some of the flags, like `-outputdir`, probably shouldn’t be duplicated
- ordering matters for a lot of the flags
- some of them conflict with one another, like `-i` erasing any previously configured load paths
- we probably don’t want to pass the module names in as that would force `ghcide` to load everything eagerly rather than when only when you actually look at the file

Maybe someday all of this will be made unnecessary by `cabal` itself, but until that happy day, this is a pretty good solution for loading & typechecking all of our sources in `ghcide`, at the cost of requiring us to do a bit more manual work around the project’s packages & versions.

PR #376, opened 2019-11-05
Expose diffGraph again

tclem:
The internal service we use to expose semantic's capabilities over http depends on `diffGraph`, so this just adds it to the module exports again.


robrix:
Whoops, thought I checked for that. My bad!

PR #385, opened 2019-11-19
JSON design doc

aymannadeem:
This design doc starts to explore how we want to produce JSON for:
1. Supporting the mobile team using the existing imprecise AST pipeline and structuring the data we get over Twirp RPC
2. Modeling how we want the new CLI in `semantic-ast` to structure its json output in terms of use-cases we may want to support

### DIFF
--- a/script/publish
+++ b/script/publish
@@ -6,7 +6,7 @@
 set -e
 cd $(dirname "$0")/..
 
-VERSION="0.8.0.0"
+VERSION="0.9.0.0"
 BUILD_SHA=$(git rev-parse HEAD 2>/dev/null)
 DOCKER_IMAGE=docker.pkg.github.com/github/semantic/semantic
 
--- a/semantic.cabal
+++ b/semantic.cabal
@@ -1,7 +1,7 @@
 cabal-version:       2.4
 
 name:                semantic
-version:             0.8.0.0
+version:             0.9.0.0
 synopsis:            Framework and executable for analyzing and diffing untrusted code.
 description:         Semantic is a library for parsing, analyzing, and comparing source code across many languages.
 homepage:            http://github.com/github/semantic#readme


semantic-ast/docs/json-ASTs.md
```diff
@@ -0,0 +1,42 @@
+**⚠️WIP:⚠️** _I'm working through the open questions listed below and will move them up into the Design section as I answer them._
+
+# JSON output for Python
+
+`semantic-ast` is dedicated to the production of ASTs. The package provides a CLI for auto-generating ASTs from source code. This is the output of the `parseByteString` utility in `haskell-tree-sitter`. Currently, we only parse Python source, but this will soon change as we add languages. In addition to supporting more languages, consumers of this library should also be able to specify the output format in which they'd like to display generated ASTs. JSON is one of such formats, and one we can prioritize given its standard usage across a variety of applications and use-cases.
```
tclem: I'd like to argue for just including this in semantic: specifically the existing `semantic parse` command and the APIs that are surfaced to semanticd. I think it's advantageous for the existing CLI to just get better AST support (more languages, precise ASTs, faster parsing) instead of expecting consumers to go find and use another utility.

semantic-ast/docs/json-ASTs.md
```diff
@@ -0,0 +1,42 @@
...
+- [Open questions](open-questions)
+
+### Use-cases
+
+Exploring how this JSON would be consumed informs the shape of the output we produce (ex., most likely one JSON object per AST node). Here are a few use-cases that come to mind:
+- External user wants to verify something about an AST, `semantic-ast` gives user an AST that can be piped into `jq` and be used to slice data in various ways
+- External user wants to process the code through _some_ research pipeline
+- GitHub action that parses your code and verifies some property (like a linter)
+- ML team: adjacency list not tree
+- Mobile team: mobile Code Nav features (though this is something to explore using the current pipeline in place, _not_ just precise ASTs)
```
tclem: As the parenthetical hints at, I think that mobile code nav should be out of scope for this document. Properly producing code nav on mobile will require implementing endpoints in the main github application so that we can properly leverage authN/Z.

semantic-ast/docs/json-ASTs.md
```diff
@@ -0,0 +1,42 @@
...
+- Us! ie., Semantic's program analysis goals
+- TBD
+
+### Possible user-facing features/endpoints
+
+1. **CLI** Currently, Semantic allows us to specify `parse`, `diff`, `graph` along with the `--json` flag to get the output we want. One open question is how closely we want to deviate from this pattern. In `semantic-ast`, we're aiming for a simpler set of options.
```
tclem: I think one of the short comings of semantic's current CLI is that formats are passed as switches. Instead it would be advantageous to have a `--format=X` style argument so that we can support a number of different formats.

semantic-ast/docs/json-ASTs.md
```diff
@@ -0,0 +1,42 @@
...
+2. **GraphQL** Exposing a GraphQL endpoint different than [the preliminary blob](https://github.com/github/github/blob/master/config/schema.internal.graphql#L9483-L9497) we provide now would be useful for bringing code nav to mobile. Historically we ran into challenges with GraphQL, since it is ideal for data the follows a predetermined structure represented by the schema. Our AST data, however, wouldn't come with this prior knowledge, and the structure would change depending on the user-specified input source code. This wouldn't really allow a user to specify an elaborate query. To get around this, we currently return a single json payload. However, for mobile, we could do something where we figure out how the schema for the data that we get over Twirp RPC in the Aleph query service (not the indexer), and use the structure most viable to the mobile team to inform how to proceed here for `semantic-ast`'s json output.
```
tclem: Mobile code nav will be relatively straightforward to implement for graphQL and doesn't have the same issues that we ran into with returning tree structures from graphQL. Again, I would vote this is out of scope for semantic-ast.

semantic-ast/docs/json-ASTs.md
aymannadeem: I was working off the vision @robrix outlined for semantic [here](https://github.com/github/semantic-code/blob/321ae2d13c7a01eee0470fde1f475e713838f7e0/architecture/README.md) in his architecture diagram and also in ongoing conversations for how he'd like us to move toward eventually deprecating the existing CLI altogether. 

I think the goals of `semantic-ast` were never really clarified anywhere though. I think discussing the future of `Semantic` and whether we want to move toward this world of smaller packages or not, and to what extent this is part of that transition probably reflects whether we subsume this stuff into Semantic's existing CLI.

semantic-ast/docs/json-ASTs.md
aymannadeem: Yes! I glommed it all in here but they are two very separate streams. This involves precise ASTs, and the mobile graphQL endpoint is more about figuring out how to structure existing (imprecise) ASTs.

semantic-ast/docs/json-ASTs.md
aymannadeem: Good call!

Additionally, I think the question of what we'd want to modify in the existing CLI fall into two broad categories: 1) the API design itself and how easy it is to work with (I found `optparse-applicative` and that entire code really complex) and 2). the CLI user experience itself.

semantic-ast/docs/json-ASTs.md
aymannadeem: This is absolutely out of scope for `semantic-ast`! I tried clarifying that but I think I should extract that entirely from this design doc. 😄 Would you be game to 🍐 tomorrow on showing me what it would take to get the graphQL endpoint in place? I was poking around Aleph docs today but could use a guide.

semantic-ast/docs/json-ASTs.md
dcreager: That doc just talks about splitting up the implementation into different packages, right? My reading of it has always been that there would still be a single top-level `semantic` package, that would primarily pull in all of the lower-level implementation packages and provide the `semantic` CLI program. That would still be the primary “UI” to access all of our functionality.  If there are any programs in any of the lower packages, they would not be installed, and would only be used for testing purposes.

PR #411, opened 2019-12-18
Update script/repl & ghcide support for 8.8.1

robrix:
This PR resolves issues with `script/repl` & `ghcide` which arose as a result of #236.


script/ghci-flags
```diff
@@ -3,63 +3,78 @@
...
-echo "-Wno-missed-specialisations"
-echo "-Wno-missing-import-lists"
-echo "-Wno-missing-local-signatures"
-echo "-Wno-monomorphism-restriction"
-echo "-Wno-name-shadowing"
-echo "-Wno-safe"
-echo "-Wno-unsafe"
-echo "-Wno-star-is-type"
+# recent hie-bios requires us to output to the file at $HIE_BIOS_OUTPUT, but older builds & script/repl don’t set that var, so we default it to stdout
+output_file="${HIE_BIOS_OUTPUT:-/dev/stdout}"
```
robrix: As @aymannadeem & I discovered, `hie-bios` (used by `ghcide`) now expects this script to write its output to a file named by `$HIE_BIOS_OUTPUT`. However, older `ghcide` versions don’t, nor does `script/repl`, and it’s inconvenient to do so if you’re trying to debug this by calling it from the shell.

We therefore default to stdout when this env var is unset.

script/ghci-flags
```diff
@@ -3,63 +3,78 @@
...
+
+# do a build of dependencies up front to ensure they’re all available
+cabal v2-build -v0 all --only-dependencies
```
robrix: Building deps up front should help avoid the need for extra reloads in VS Code.

script/ghci-flags
```diff
@@ -3,63 +3,78 @@
...
+
+build_products_dir="dist-newstyle/build/x86_64-osx/ghc-$ghc_version/build-repl"
```
robrix: 1. We only use this directory for `script/repl` since `ghcide` doesn’t do codegen.

2. There’s no reasonable way for us to use the same set of per-component directories as `cabal` will for build products since we can’t reasonably customize this flag based on e.g. the source path.

This means that we’re better off picking a path that `cabal` never will, in order to avoid `cabal build` and `script/repl` attempting to use each others’ interface & object files (cf https://github.com/haskell/cabal/issues/3565 & https://github.com/haskell/cabal/issues/6059) if we e.g. attempt to run `cabal test` from inside `ghci`.

That in turn obviates the need for us to either track the package’s name and version for the path, or to support the use of a non-standard build directory for `script/repl`.

(**NB:** `hie-bios` apparently calls this script with the path to each specific source file as they’re loaded, so we should in principle be able to map those paths back onto specific sets of flags! But `ghci` doesn’t, and since `ghcide` doesn’t build, there’d be no point.)

script/ghci-flags
```diff
@@ -3,63 +3,78 @@
...
-echo "-i$root/semantic-ast/src"
-echo "-i$root/semantic-core/src"
-echo "-i$root/semantic-java/src"
-echo "-i$root/semantic-json/src"
-echo "-i$root/semantic-python/src"
-echo "-i$root/semantic-tags/src"
-echo "-i$root/app"
-echo "-i$root/src"
-echo "-i$root/bench"
-echo "-i$root/test"
```
robrix: We no longer need to use absolute paths for these settings as whatever VS Code or `ghcide` bug it was which required this has now been fixed.

script/ghci-flags
```diff
@@ -3,63 +3,78 @@
...
+  echo "-Wno-implicit-prelude"
+  echo "-Wno-missed-specialisations"
+  echo "-Wno-missing-import-lists"
+  echo "-Wno-missing-local-signatures"
+  echo "-Wno-monomorphism-restriction"
+  echo "-Wno-name-shadowing"
+  echo "-Wno-safe"
+  echo "-Wno-unsafe"
+  [[ "$ghc_version" = 8.6.* ]] || [[ "$ghc_version" = 8.8.* ]] && echo "-Wno-star-is-type" || true
+  [[ "$ghc_version" = 8.8.* ]] && echo "-Wno-missing-deriving-strategies" || true
```
robrix: `semantic` only supports a single version of `ghc`, but this minimizes the amount of editing necessary when copying it to a new project.

script/repl
```diff
@@ -4,12 +4,7 @@
 
 set -e
 
-cd $(dirname "$0")/..
+cd "$(dirname "$0")/.."
 
-repl_builddir=dist-repl
-
-if [[ ! -d $repl_builddir ]]; then
-  echo "$repl_builddir does not exist, first run 'cabal repl --builddir=$repl_builddir', exit, and then re-run $0"
```
robrix: This is no longer necessary, but it _might_ be necessary to do a `cabal build` (of at least part of the project) to populate the package db. However, in those cases the error messages you see inform you of that fact already so there isn’t much benefit to adding an extra check.

PR #414, opened 2019-12-19
More precise parsing

tclem:
Enables precise ASTs for:

- [x] Go
- [x] TypeScript
- [x] TSX (and JSX and JavaScript)


semantic-go/src/Language/Go/Tags.hs
```diff
@@ -0,0 +1,95 @@
+{-# LANGUAGE AllowAmbiguousTypes, DataKinds, DisambiguateRecordFields, FlexibleContexts, FlexibleInstances, MultiParamTypeClasses, NamedFieldPuns, PartialTypeSignatures, ScopedTypeVariables, TypeApplications, TypeFamilies, TypeOperators, UndecidableInstances #-}
```
aymannadeem: can we use the alternate style with one pragma per extension, each on their own line?

tclem: Yes, that's a very good idea. I'll fix that for at least the new files I've added on this PR. 👍

semantic-tsx/src/Language/TSX.hs
```diff
@@ -0,0 +1,21 @@
+{-# OPTIONS_GHC -freduction-depth=0 #-}
```
tclem: @robrix - OK to leave this in for now? Or...

test/Examples.hs
```diff
@@ -78,17 +103,51 @@ rubySkips = Path.relFile <$>
...
 
   -- Can't detect method calls inside heredoc bodies with precise ASTs
   , "ruby_spec/core/argf/readpartial_spec.rb"
   , "ruby_spec/core/process/exec_spec.rb"
   ]
 
+tsxSkips :: [Path.RelFile]
+tsxSkips = Path.relFile <$>
+  [
+  -- Cannot decode byte '\xe2': Data.Text.Internal.Encoding.decodeUtf8: Invalid UTF-8 stream
```
tclem: Almost all languages have this same kind of encoding problem with certain example files that contain non-utf8 characters. Should we just leave this as a known issue for now?

test/fixtures/go/corpus/assignment-statements.diffA-B.txt
```diff
@@ -8,12 +8,7 @@
       { (Identifier)
       ->(Identifier) }
         (Integer))
-      (Assignment
-        (Statements
-        { (Identifier)
-        ->(Identifier) }
-        { (Identifier)
-        ->(Identifier) })
+      (AugmentedAssignment
```
tclem: Various changes to shape of assignment trees due to things like adding `AugmentedAssignment`.

semantic-tsx/src/Language/TSX.hs
robrix: 👍

test/Examples.hs
robrix: cc @maxbrunsfeld

maxbrunsfeld: What is the error? Tree-sitter *should* just treat this as a syntax error; it shouldn't halt the whole thing.

tclem: `Cannot decode byte '\xe2': Data.Text.Internal.Encoding.decodeUtf8: Invalid UTF-8 stream` is the error, it's coming from [unmarshalling](https://github.com/tree-sitter/haskell-tree-sitter/blob/master/tree-sitter/src/TreeSitter/Unmarshal.hs#L124) in haskell-tree-sitter.

maxbrunsfeld: Oh, I see. @robrix @aymannadeem For that Text-unmarshalling step, can we do some kind of *lossy* UTF8 decoding, e.g. in which invalid byte sequences are replaced by the [replacement character](https://en.wikipedia.org/wiki/Specials_(Unicode_block)#Replacement_character). I don't think we want to bail out of analyzing the whole file if there are some non-utf8 bytes somewhere.

tclem: Yes!

``` haskell
decodeUtf8With lenientDecode
```

robrix: Oooh, my bad, I misread the error. Fortunately @tclem fixed this yesterday!

PR #416, opened 2020-01-13
Higher-order Traversable

robrix:
This PR replaces the `GFoldable1` class with a `Traversable1` class, its generic counterpart, and a `Generics` newtype suitable for bridging the two until such time as we derive `Traversable1` instances for our TH-generated AST datatypes.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -53,7 +53,7 @@ jobs:
 
     - name: hlint
       run: |
-        cabal install hlint --installdir=dist-newstyle
+        cabal install hlint --installdir=dist-newstyle --overwrite-policy=always
         dist-newstyle/hlint src semantic-python
 
     - name: Build & test
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -53,7 +53,7 @@ jobs:
 
     - name: hlint
       run: |
-        cabal install hlint --installdir=dist-newstyle --overwrite-policy=always
+        cabal install hlint --installdir=dist-newstyle
         dist-newstyle/hlint src semantic-python
 
     - name: Build & test
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -53,7 +53,7 @@ jobs:
 
     - name: hlint
       run: |
-        cabal install hlint --installdir=dist-newstyle
+        test -f dist-newstyle/hlint || cabal install hlint --installdir=dist-newstyle
         dist-newstyle/hlint src semantic-python
 
     - name: Build & test
--- a/semantic-tags/src/Tags/Tagging/Precise.hs
+++ b/semantic-tags/src/Tags/Tagging/Precise.hs
@@ -1,12 +1,12 @@
-{-# LANGUAGE AllowAmbiguousTypes   #-}
-{-# LANGUAGE ConstraintKinds       #-}
-{-# LANGUAGE FlexibleContexts      #-}
-{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE AllowAmbiguousTypes #-}
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE RankNTypes            #-}
-{-# LANGUAGE ScopedTypeVariables   #-}
-{-# LANGUAGE TypeApplications      #-}
-{-# LANGUAGE TypeOperators         #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
 module Tags.Tagging.Precise
 ( Tags
 , ToTags(..)
--- a/semantic-tags/src/Tags/Tagging/Precise.hs
+++ b/semantic-tags/src/Tags/Tagging/Precise.hs
@@ -6,6 +6,7 @@ module Tags.Tagging.Precise
 , runTagging
 , firstLine
 , GFoldable1(..)
+, GTraversable1(..)
 ) where
 
 import Control.Carrier.Reader
@@ -76,3 +77,6 @@ instance (Foldable f, GFoldable1 c g) => GFoldable1 c (f :.: g) where
 
 instance GFoldable1 c U1 where
   gfoldMap1 _ _ = mempty
+
+
+class GTraversable1 c t where
--- a/semantic-tags/src/Tags/Tagging/Precise.hs
+++ b/semantic-tags/src/Tags/Tagging/Precise.hs
@@ -97,6 +97,8 @@ instance GFoldable1 c U1 where
 
 -- FIXME: move Traversable1 into semantic-ast.
 -- FIXME: derive Traversable1 instances for TH-generated syntax types.
+
+-- | Simultaneous traversal of subterms of kind @*@ and @* -> *@ in an 'Applicative' context.
 class Traversable1 c t where
   -- | Map annotations and subterms of kind @* -> *@ into an 'Applicative' context.
   traverse1
--- a/semantic-tags/src/Tags/Tagging/Precise.hs
+++ b/semantic-tags/src/Tags/Tagging/Precise.hs
@@ -1,5 +1,6 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE DefaultSignatures #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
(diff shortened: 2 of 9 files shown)


semantic-tags/src/Tags/Tagging/Precise.hs
```diff
@@ -43,36 +64,133 @@ firstLine :: Source -> Text
...
 
 
--- FIXME: move GFoldable1 into semantic-ast.
-class GFoldable1 c t where
-  -- | Generically map functions over fields of kind @* -> *@, monoidally combining the results.
-  gfoldMap1
-    :: Monoid b
-    => (forall f . c f => f a -> b)
+-- FIXME: move Traversable1 into semantic-ast.
+-- FIXME: derive Traversable1 instances for TH-generated syntax types.
```
robrix: If `semantic-ast` were a published hackage package along the lines of `semantic-source`, then we could move `Traversable1` & friends into it, the `tree-sitter-*` packages could depend on it, and we could derive `Traversable1` instances for all of the generated syntax types.

On the other hand, the same would work if we defined this in some _other_ package. `semantic-ast` just seems like a reasonable one to choose.

semantic-tags/src/Tags/Tagging/Precise.hs
```diff
@@ -43,36 +64,133 @@ firstLine :: Source -> Text
...
+
+-- | Simultaneous traversal of subterms of kind @*@ and @* -> *@ in an 'Applicative' context.
+--
+-- 'Traversable1' can express any combination of first- and second-order mapping, folding, and traversal.
+class Traversable1 c t where
```
robrix: I’m really quite pleased with this class. It:

1. Subsumes the prior `GFoldable1` functionality.
2. Can express `Foldable` (when applied recursively).
3. Can express `Functor` (when applied recursively).
4. Can express `Traversable` (when applied recursively).
5. Suffices to express all of the tagging traversals.

The only bit I don’t really love is the name. There are two distinct uses of the suffix `1` in `base` that I’m aware of:

1. Classes such as `Generic1` & `Show1`, where it indicates that a similar class applied to kind `*` has been lifted to kind `* -> *`. This usage is most analogous to how we’re using it for `Traversable1`, and `base` does not define `Traversable1`, `Functor1`, or `Foldable1`.

2. Functions such as `foldl1` and `foldr1`, where it indicates that the function expects a non-empty collection. This use is further applied in the widely-depended-upon (directly or indirectly) [`semigroupoids` package](http://hackage.haskell.org/package/semigroupoids), which supplies `Foldable1` and `Traversable1` classes modelling non-empty foldable and traversable containers. `Data.List.NonEmpty` was moved into `base`, and it seems plausible that these might eventually be as well, which makes me a bit reluctant to use this name.

On the other hand, in `fused-effects` and `fused-syntax` we use the prefix `H` to denote higher-order analogues of the un-prefixed name, e.g. `HFunctor` is analogous to `Functor` but for `((* -> *) -> (* -> *)`. But the meaning here is different, too: the subterms are all of a homogeneous parameter type, not heterogeneous un-parameterized (constant) types. Our `Traversable1` cannot change the types of subterms, only of their annotations, which means that it’s a strictly different usage, somewhere between `H*` and the [`mono-traversable` package](http://hackage.haskell.org/package/mono-traversable).

If we do call it `Traversable1`, and publish it, and so forth, I suppose the worst that can happen is that we might occasionally have to qualify which one’s name we mean—I don’t count it as terribly likely that `Foldable1` or `Traversable1` will ever end up in `base`. So maybe this name is ok, and I’m just overthinking it?

@github/semantic-code: I would appreciate your thoughts.

semantic-tags/src/Tags/Tagging/Precise.hs
```diff
@@ -43,36 +64,133 @@ firstLine :: Source -> Text
...
+  traverse1 f g = fmap to1 . gtraverse1 @c f g . from1
+
+for1
+  :: forall c t f a b
+  .  (Traversable1 c t, Applicative f)
+  => t a
+  -> (a -> f b)
+  -> (forall t' . c t' => t' a -> f (t' b))
+  -> f (t b)
+for1 t f g = traverse1 @c f g t
```
robrix: I’ve also supplied some of the standard `Traversable`-family functions, lifted to `Traversable1`.

semantic-tags/src/Tags/Tagging/Precise.hs
```diff
@@ -43,36 +64,133 @@ firstLine :: Source -> Text
...
+  :: forall c t f a a' a''
+  .  (Traversable1 c t, Applicative f)
+  => t a
+  -> (a -> f a')
+  -> (forall t' . c t' => t' a -> f a'')
+  -> f ()
+for1_ t f g = getAp $ foldMap1 @c (Ap . void . f) (Ap . void . g) t
+
+foldMap1 :: forall c t b a . (Traversable1 c t, Monoid b) => (a -> b) -> (forall t' . c t' => t' a -> b) -> t a -> b
+foldMap1 f g = getConst . traverse1 @c (Const . f) (Const . g)
```
robrix: I’ve removed `GFoldable1` instead of defining `Foldable1` and making it a superclass of `Traversable1` largely because we are focusing our attention on the ASTs we get out of TH. We might eventually decide that this is useful enough to warrant some more interesting hierarchy, but I kind of doubt that there are many cases where we’d really have non-`Traversable1` `Foldable1` types.

semantic-tags/src/Tags/Tagging/Precise.hs
```diff
@@ -43,36 +64,133 @@ firstLine :: Source -> Text
...
+foldMapDefault1 :: (Traversable1 Foldable t, Monoid b) => (a -> b) -> t a -> b
+foldMapDefault1 f = foldMap1 @Foldable f (foldMap f)
+
+-- | This function may be used as a value for 'fmap' in a 'Functor' instance.
+fmapDefault1 :: Traversable1 Functor t => (a -> b) -> t a -> t b
+fmapDefault1 f = runIdentity . traverse1 @Functor (Identity . f) (Identity . fmap f)
+
+-- | This function may be used as a value for 'traverse' in a 'Traversable' instance.
+traverseDefault1 :: (Traversable1 Traversable t, Applicative f) => (a -> f b) -> t a -> f (t b)
+traverseDefault1 f = traverse1 @Traversable f (traverse f)
```
robrix: These are analogous to `foldMapDefault` and `fmapDefault` from `base`, and extend even to `traverse`. Note the `Foldable`, `Functor`, and `Traversable` constraints passed to `Traversable1`—`traverse1` does a shallow traversal of the type, so the recursive traversal is performed by requiring these constraints to be present on subterms.

semantic-tags/src/Tags/Tagging/Precise.hs
```diff
@@ -43,36 +64,133 @@ firstLine :: Source -> Text
...
+
+-- FIXME: move GTraversable1 into semantic-ast.
+class GTraversable1 c t where
+  -- | Generically map annotations and subterms of kind @* -> *@ into an 'Applicative' context.
+  gtraverse1
+    :: Applicative f
+    => (a -> f b)
+    -> (forall t' . c t' => t' a -> f (t' b))
+    -> t a
+    -> f (t b)
```
robrix: This is fairly boring IMO.

semantic-tags/src/Tags/Tagging/Precise.hs
```diff
@@ -43,36 +64,133 @@ firstLine :: Source -> Text
...
-instance GFoldable1 c (K1 R t) where
-  gfoldMap1 _ _ = mempty
+-- | @'Generics' t@ has a 'Traversable1' instance when @'Rep1' t@ has a 'GTraversable1' instance, making this convenient for applying 'traverse1' to 'Generic1' types lacking 'Traversable1' instances:
+--
+-- @
+-- 'getGenerics' '<$>' 'traverse1' f g ('Generics' t) = 'to1' '<$>' 'gtraverse1' f g ('from1' t)
+-- @
+--
+-- It further defines its 'Foldable', 'Functor', and 'Traversable' instances using 'Traversable1', making it suitable for deriving with @-XDerivingVia@.
+newtype Generics t a = Generics { getGenerics :: t a }
```
robrix: This, on the other hand, is kind of fun.

We don’t have `Traversable1` instances for our AST types yet, so this bridges the gap—wrap a `Generic1` term in `Generics` and you can use `Traversable1` by means of said instance.

It also supports use of `-XDerivingVia` (for classes supporting `-XGeneralizedNewtypeDeriving`… alas, this does not include `Traversable` or `Traversable1`), which is what prompted the name. I intend to measure how long it takes to derive `Foldable` & `Functor` using `-XDeriveFoldable`/`-XDeriveFunctor`, vs. how long it takes using `-XDerivingVia` for our TH-generated AST datatypes:

```haskell
newtype Expr a = Expr ((Thing :+: OtherThing) a)
  deriving (Foldable, Functor) via Generics
```

.github/workflows/haskell.yml
```diff
@@ -53,7 +53,7 @@ jobs:
 
     - name: hlint
       run: |
-        cabal install hlint --installdir=dist-newstyle
+        test -f dist-newstyle/hlint || cabal install hlint --installdir=dist-newstyle
```
robrix: I was getting CI failures because this already existed. I opted not to overwrite it in that case and test for the existence of the binary before installing instead.

semantic-tags/src/Tags/Tagging/Precise.hs
joshvera: I think `Traversable1` is fine if we keep it in these functions. Maybe a note in the comment explaining that this traversale of subterms is unique in that it can affect annotations?

PR #420, opened 2020-01-15
Non-overlapping ToTags instances

robrix:
🔥 the advanced overlap in the various `.Tags` modules, improving compile times by a large factor (~12s to ~3s for TSX, say) and obviating the need for `-freduction-depth=0`.

Adds a bunch of `instance ToTags Lang.Blah` lines but so be it?

### DIFF
--- a/semantic-go/src/Language/Go/Tags.hs
+++ b/semantic-go/src/Language/Go/Tags.hs
@@ -1,17 +1,9 @@
-{-# LANGUAGE AllowAmbiguousTypes      #-}
-{-# LANGUAGE DataKinds                #-}
+{-# LANGUAGE DefaultSignatures #-}
 {-# LANGUAGE DisambiguateRecordFields #-}
-{-# LANGUAGE FlexibleContexts         #-}
-{-# LANGUAGE FlexibleInstances        #-}
-{-# LANGUAGE MultiParamTypeClasses    #-}
-{-# LANGUAGE NamedFieldPuns           #-}
-{-# LANGUAGE PartialTypeSignatures    #-}
-{-# LANGUAGE ScopedTypeVariables      #-}
-{-# LANGUAGE TypeApplications         #-}
-{-# LANGUAGE TypeFamilies             #-}
-{-# LANGUAGE TypeOperators            #-}
-{-# LANGUAGE UndecidableInstances     #-}
-
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
 module Language.Go.Tags
 ( ToTags(..)
 ) where
@@ -26,6 +18,7 @@ import           Source.Source as Source
 import           Tags.Tag
 import qualified Tags.Tagging.Precise as Tags
 import qualified TreeSitter.Go.AST as Go
+import           TreeSitter.Token
 
 class ToTags t where
   tags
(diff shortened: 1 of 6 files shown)


semantic-go/src/Language/Go/Tags.hs
```diff
@@ -34,58 +27,47 @@ class ToTags t where
...
-
-type family ToTagsInstance t :: Strategy where
-  ToTagsInstance (_ :+: _)              = 'Custom
-  ToTagsInstance Go.FunctionDeclaration = 'Custom
-  ToTagsInstance Go.MethodDeclaration   = 'Custom
-  ToTagsInstance Go.CallExpression      = 'Custom
-  ToTagsInstance _                      = 'Generic
-
-instance ToTagsBy 'Custom Go.FunctionDeclaration where
-  tags' t@Go.FunctionDeclaration
```
robrix: Instead of defining three classes and a type family to select whether we want to use the generic or a custom implementation for each type…

semantic-go/src/Language/Go/Tags.hs
```diff
@@ -34,58 +27,47 @@ class ToTags t where
...
+instance ToTags Go.FunctionDeclaration where
+  tags t@Go.FunctionDeclaration
```
robrix: …we specify normal instances for the types we want to customize…

semantic-go/src/Language/Go/Tags.hs
```diff
@@ -34,58 +27,47 @@ class ToTags t where
...
-  tags'
+  default tags
     :: ( Has (Reader Source) sig m
        , Has (Writer Tags.Tags) sig m
+       , Generic1 t
+       , Tags.GTraversable1 ToTags (Rep1 t)
        )
     => t Loc
     -> m ()
+  tags = gtags
```
robrix: …with `DefaultSignatures` providing the generic behaviour…

semantic-go/src/Language/Go/Tags.hs
```diff
@@ -97,11 +79,108 @@ gtags
...
   Tags.yield (Tag name kind loc (Tags.firstLine sliced) Nothing)
+
+
+instance ToTags Go.ArgumentList
+instance ToTags Go.ArrayType
+instance ToTags Go.AssignmentStatement
+instance ToTags Go.BinaryExpression
+instance ToTags Go.BlankIdentifier
+instance ToTags Go.Block
+instance ToTags Go.BreakStatement
```
robrix: …for the explicitly-listed remainder of the types.

semantic-go/src/Language/Go/Tags.hs
```diff
@@ -97,11 +79,108 @@ gtags
...
+instance ToTags Go.FallthroughStatement
+instance ToTags Go.False
+instance ToTags Go.FieldDeclaration
+instance ToTags Go.FieldDeclarationList
+instance ToTags Go.FieldIdentifier
+instance ToTags Go.FloatLiteral
+instance ToTags Go.ForClause
+instance ToTags Go.ForStatement
+instance ToTags Go.FuncLiteral
+-- instance ToTags Go.FunctionDeclaration
```
robrix: I’ve left commented-out lines for the customized instances in the alphabetized lists. It’s not super important but it provides a nice overview.

semantic-python/src/Language/Python/Tags.hs
```diff
@@ -1,26 +1,19 @@
...
-{-# LANGUAGE TypeApplications         #-}
-{-# LANGUAGE TypeFamilies             #-}
-{-# LANGUAGE TypeOperators            #-}
-{-# LANGUAGE UndecidableInstances     #-}
-
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
```
robrix: One side-effect is that we can get away with far fewer language extensions now.

semantic-tsx/src/Language/TSX/Tags.hs
```diff
@@ -1,19 +1,10 @@
...
-{-# LANGUAGE NamedFieldPuns           #-}
-{-# LANGUAGE OverloadedStrings        #-}
-{-# LANGUAGE PartialTypeSignatures    #-}
-{-# LANGUAGE ScopedTypeVariables      #-}
-{-# LANGUAGE TypeApplications         #-}
-{-# LANGUAGE TypeFamilies             #-}
-{-# LANGUAGE TypeOperators            #-}
-{-# LANGUAGE UndecidableInstances     #-}
-
-{-# OPTIONS_GHC -freduction-depth=0 #-}
```
robrix: Another, more important effect is that we no longer require `-freduction-depth=0` for TSX or TypeScript simply because there’s so much less work for the typechecker to do. This also makes them compile much more quickly!

PR #435, opened 2020-01-21
Unmarshal timeout

tclem:
We have a timeout that gets passed to tree-sitter as a way to be able to abort parsing across the FFI (important if a parser runs away for some reason or to maintain some other service objective), but with the move away from assignment we no longer use the haskell `System.timeout` to protect the non C aspects of producing an AST. This adds a new config setting that allows us to set a timeout on unmarshalling.


tclem:
This also includes the unbalanced tree and interruptible fixes from https://github.com/tree-sitter/haskell-tree-sitter/pull/262.

PR #444, opened 2020-01-27
replace auto-derived ToJSON instance with one that mirrors ToJSONFields

aymannadeem:


semantic-source/src/Source/Range.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE DeriveAnyClass, DeriveGeneric, RankNTypes #-}
+{-# LANGUAGE DeriveAnyClass, DeriveGeneric, RankNTypes, NamedFieldPuns, OverloadedStrings #-}
```
robrix: Is `OverloadedStrings` necessary here?

semantic-source/src/Source/Loc.hs
```diff
@@ -13,6 +13,7 @@ import GHC.Generics (Generic)
 import Prelude hiding (span)
 import Source.Range
 import Source.Span
+import Data.Aeson (ToJSON(..), object, (.=))
```
robrix: Could you please alphabetize the imports? 🙏

aymannadeem: yep my bad!

semantic-source/src/Source/Range.hs
aymannadeem: Removed.

PR #469, opened 2020-02-04
Traversable1 instances for AST datatypes

robrix:
This PR:

- [x] Moves `AST.Element` from `semantic-tags` to `semantic-ast`, where it belongs.
- [x] Moves `Traversable1` into `semantic-ast`.
- [x] Defines `Traversable1` instances for AST datatypes.
- [x] Defines `Foldable`, `Functor`, and `Traversable` instances for AST datatypes in terms of their derived `Traversable1` instances. This cuts something on the order of six seconds off compiles of `Language.Python.AST` and should be more significant for larger languages’ AST modules.
- [x] Implements tagging using `Traversable1` instances instead of `GTraversable1`. This should significantly reduce the amount of work GHC has to do to compile the tagging modules, but I haven’t measured the delta in compile time.

NB: It might be possible for scope graphing to take advantage of `Traversable1` now, but I haven’t specifically looked at that. cc @joshvera @patrickt


semantic-ast/semantic-ast.cabal
```diff
@@ -37,7 +37,10 @@ common haskell
 
 library
   import:              haskell
-  exposed-modules:     Marshal.JSON
+  exposed-modules:
+    AST.Element
+    AST.Traversable1
+    Marshal.JSON
```
robrix: 1. `AST.Element` was moved here from `semantic-tags`.
2. `AST.Traversable1` is a new module, but its contents were likewise moved from `semantic-tags`.
3. We should rename `Marshal.JSON` to `AST.Marshal.JSON` or something like that, but I haven’t done that in this PR. cc @aymannadeem

semantic-ast/src/AST/Traversable1.hs
```diff
@@ -0,0 +1,158 @@
...
+{-# LANGUAGE DefaultSignatures #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
+module AST.Traversable1
```
robrix: Only delta in this module is the removal of a few, now obsolete, `FIXME`s 🎉

semantic-codegen/src/AST/GenerateSyntax.hs
```diff
@@ -71,34 +72,55 @@ syntaxDatatype language allSymbols datatype = skipDefined $ do
...
 
+makeTraversalInstances :: TypeQ -> Q [Dec]
+makeTraversalInstances ty =
+  [d|
+  instance Foldable $ty where
+    foldMap = foldMapDefault1
+  instance Functor $ty where
+    fmap = fmapDefault1
+  instance Traversable $ty where
+    traverse = traverseDefault1
```
robrix: These instances compile significantly faster than the derived ones did, which more than makes up for the addition of the new `Traversable1` instance.

semantic-go/src/Language/Go/AST.hs
```diff
@@ -18,4 +18,4 @@ import           Prelude hiding (False, Float, Integer, Rational, String, True)
 import           AST.GenerateSyntax
 import qualified Language.Go.Grammar as Grammar
 
-astDeclarationsForLanguage Grammar.tree_sitter_go "../../../vendor/tree-sitter-go/src/node-types.json"
\ No newline at end of file
+astDeclarationsForLanguage Grammar.tree_sitter_go "../../../vendor/tree-sitter-go/src/node-types.json"
```
robrix: @aymannadeem didn’t catch this earlier, but you should probably configure your editor to insert newlines at EOF. (I’m actually surprised it doesn’t already.)

semantic-python/semantic-python.cabal
```diff
@@ -25,6 +25,7 @@ common haskell
                      , fused-syntax
                      , parsers ^>= 0.12.10
                      , semantic-analysis ^>= 0
+                     , semantic-ast
```
robrix: Note that this does not, as far as I’m aware, present an obstacle due to per-component builds; `semantic-ast`’s executable component depends on `semantic-python`, but the dependency graph is still acyclic since its library component does not.

semantic-ruby/src/Language/Ruby/Tags.hs
```diff
@@ -89,7 +88,7 @@ instance ToTags Rb.Class where
     where
       range' = case extraChildren of
         Prj Rb.Superclass { ann = Loc { byteRange = Range { end }}} : _ -> Range start end
-        _ -> Range start (getEnd expr)
+        _                                                               -> Range start (getEnd expr)
```
robrix: Stuff like this is just `stylish-haskell` having fun.

semantic-ast/semantic-ast.cabal
```diff
@@ -49,33 +53,10 @@ library
...
                      , bytestring           ^>= 0.10.8.2
                      , optparse-applicative >= 0.14.3 && < 0.16
                      , pretty-simple        ^>= 3.1.0.0
                      , text                 ^>= 1.2.3.1
 
   hs-source-dirs:      src
   default-language:    Haskell2010
-
-
-executable semantic-ast
```
robrix: This has been moved to `semantic-parse`.

semantic-ast/src/AST/Traversable1/Class.hs
```diff
@@ -0,0 +1,100 @@
...
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE UndecidableInstances #-}
+-- | This module defines the 'Traversable1' class and its generic derivation using 'GTraversable1'. Note that any changes to this file will require recompilation of all of the AST modules, which is quite expensive; thus, most additions should be made in "AST.Traversable1" instead, and that that module should not be imported by the AST modules.
+module AST.Traversable1.Class
```
robrix: Splitting the `Traversable1`-relevant definitions up between `AST.Traversable1` & `AST.Traversable1.Class` like this means that `AST.GenerateSyntax`, and thus the lang-specific AST modules, can import this module, which should hopefully be very stable, while changes can be made to the other one. Consumers of the API should use the `AST.Traversable1` module, which re-exports this one and adds a few new definitions.

This is intended to minimize expensive rebuilding of the AST modules.

semantic-python/semantic-python.cabal
robrix: In order to work around this, I have moved `semantic-ast`’s executable into the new `semantic-parse` package, which can safely depend on the lang-specific packages.

semantic-ast/semantic-ast.cabal
aymannadeem: added to my todos in #468

PR #497, opened 2020-02-21
Introduce the beginning of the scope graph API

tclem:
Implements the external API plumbing to support producing incremental scope graphs for blobs.


proto/semantic.proto
```diff
@@ -174,3 +178,30 @@ message Span {
...
+  repeated ScopeGraphNode nodes = 3;
+  repeated ScopeGraphPath paths = 4;
+  repeated ParseError errors = 5;
+}
+
+message ScopeGraphNode {
+  int32 id = 1;
+  string name = 2;
+  string line = 3;
+  string syntax = 4;
```
patrickt: What does `syntax` mean in this context? Does this describe the node type (push, pop, etc?) If so I imagine we’ll do some parsing when converting to the native data type.

proto/semantic.proto
```diff
@@ -174,3 +178,30 @@ message Span {
...
+  bool is_definition = 5;
+  Span span = 6;
+}
+
+message ScopeGraphPath {
+  string starting_symbol_stack = 1;
```
patrickt: These will be encoded values too, right? Colon-separated or something?

proto/semantic.proto
```diff
@@ -174,3 +178,30 @@ message Span {
...
+message ScopeGraphFile {
+  string path = 1;
+  string language = 2;
+  repeated ScopeGraphNode nodes = 3;
+  repeated ScopeGraphPath paths = 4;
+  repeated ParseError errors = 5;
+}
+
+message ScopeGraphNode {
+  int32 id = 1;
```
patrickt: FYI Haskell native `Int` is 64-bits wide, so we could have slightly more efficient conversion (no `fromIntegral` upcasting) if we went with `int64`.

src/Semantic/Api/ScopeGraph.hs
```diff
@@ -0,0 +1,125 @@
...
+                   , Has Parse sig m, Traversable t
+                   )
+  => t Blob
+  -> m ParseTreeScopeGraphResponse
+parseScopeGraph blobs = do
+  terms <- distributeFor blobs go
+  pure $ defMessage & P.files .~ toList terms
+  where
+    go :: (Has (Error SomeException) sig m, Has Parse sig m) => Blob -> m ScopeGraphFile
+    go blob = catching $ graphToFile <$> graphForBlob blob
```
patrickt: Really elegant definition 👍

proto/semantic.proto
```diff
@@ -174,3 +178,30 @@ message Span {
   Position start = 1;
   Position end = 2;
 }
+
+message ScopeGraphFile {
```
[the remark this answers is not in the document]
joshvera: 👍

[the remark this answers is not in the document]
tclem: :+1: let me rename these all to `StackGraph*`

proto/semantic.proto
tclem: I'm going to rename this to `kind` (which is what we call this in the other proto responses) - It's essential the name of the syntax term (`Function`, `Call`, etc).

proto/semantic.proto
tclem: Yes, these are currently stacks encoded and using ascii delimiters.

dcreager: Would it make more sense to have this be a `repeated string` instead?  Are we using different ascii characters as different delimiters, or as sigils at the beginning of a symbol name to indicate its type?  If the former, this is worth a comment describing the format more precisely since it becomes part of the API contract.  If the latter that would line up well with `repeated string`, and then only aleph would need to know specifically how we turn the list of symbols on the stack into a scalar string in the db.  And the per-language stack graph generators would have full flexibility to choose sigils and whatnot as needed for the language.

src/Semantic/Api/ScopeGraph.hs
```diff
@@ -0,0 +1,37 @@
...
+import           Proto.Semantic_JSON ()
+import           Semantic.Api.Bridge
+import qualified Semantic.Api.LegacyTypes as Legacy
+import           Semantic.Config
+import           Semantic.Task
+import           Serializing.Format (Format)
+import           Source.Loc as Loc
+
+parseScopeGraph :: ( Has Distribute sig m
+                   , Has (Error SomeException) sig m
```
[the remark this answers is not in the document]
tclem: This attempting to protect from _any_ error that might occur in parsing and generating a graph for a single file. One file failing shouldn't bring down the entire response, instead we return the rest of the results and an error placeholder for the failed one.

src/Semantic/Api/ScopeGraph.hs
```diff
@@ -0,0 +1,37 @@
...
+import           Proto.Semantic_Fields as P
+import           Proto.Semantic_JSON ()
+import           Semantic.Api.Bridge
+import qualified Semantic.Api.LegacyTypes as Legacy
+import           Semantic.Config
+import           Semantic.Task
+import           Serializing.Format (Format)
+import           Source.Loc as Loc
+
+parseScopeGraph :: ( Has Distribute sig m
```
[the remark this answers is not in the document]
tclem: Left this alone for now as I found it tricky to mix and match task, effects, and things just in IO. This matches existing handling in existing api endpoints, and we've scope threaded to see green thread behavior on this a couple of times.

PR #517, opened 2020-03-13
Detect and tag functions in variable_declarators

tclem:
Picks up a few cases where `variable_declarators` define functions in JavaScript/TypeScript and includes those in the symbols/tags output.

### DIFF
--- a/semantic-tsx/src/Language/TSX/Tags.hs
+++ b/semantic-tsx/src/Language/TSX/Tags.hs
@@ -112,6 +112,19 @@ instance ToTags Tsx.Module where
         _                           -> gtags t
       yield text = yieldTag text Module loc byteRange >> gtags t
 
+instance ToTags Tsx.VariableDeclarator where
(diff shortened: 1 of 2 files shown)


semantic-tsx/src/Language/TSX/Tags.hs
```diff
@@ -295,7 +308,7 @@ instance ToTags Tsx.Undefined
 instance ToTags Tsx.UnionType
 instance ToTags Tsx.UpdateExpression
 instance ToTags Tsx.VariableDeclaration
-instance ToTags Tsx.VariableDeclarator
+-- instance ToTags Tsx.VariableDeclarator
```
rewinfrey: Looks like we can 🔥 this now

tclem: The pattern I've been using is to comment them out since generating the full list to begin with was automated and pasted in. @robrix do you have an opinion here?

PR #528, opened 2020-04-02
Tag `QualifiedRhs` for additional references

rewinfrey:
This patch enables additional references for `CodeQL`. Thank you for the feedback @alexet 🙇 

Previously, a reference in the form of:

```codeql
from VarargsFunction f, FunctionCall fc, string terminator, int cnt
where
  terminator = f.normalTerminator(cnt) and
  ...
```
Would not extract `normalTerminator` as a reference. This is now fixed. [original source](https://github.com/Semmle/ql/blob/604731ba6b06b43e31c9891d1471bcd8b9dce936/cpp/ql/src/Security/CWE/CWE-121/UnterminatedVarargsCall.ql#L77)


semantic-codeql/src/Language/CodeQL/Tags.hs
```diff
@@ -100,19 +100,22 @@ instance ToTags CodeQL.DatatypeBranch where
     , name = CodeQL.ClassName { text }
     } = yieldTag text Class loc byteRange >> gtags t
 
-instance ToTags CodeQL.HigherOrderTerm where
-  tags t@CodeQL.HigherOrderTerm
-    { ann = loc@Loc { byteRange }
-    , name = CodeQL.LiteralId { text }
-    } = yieldTag text Call loc byteRange >> gtags t
-
```
rewinfrey, line 108: @alexet noted this `ToTags` instance is unnecessary as it always resolves to special built-in QL literals.

semantic-codeql/src/Language/CodeQL/Tags.hs
```diff
@@ -100,19 +100,22 @@ instance ToTags CodeQL.DatatypeBranch where
...
 
+instance ToTags CodeQL.QualifiedRhs where
+  tags t@CodeQL.QualifiedRhs
+    { ann = loc@Loc { byteRange }
+    , name = expr
+    } = case expr of
+          Just (Prj CodeQL.PredicateName { text }) -> yieldTag text Call loc byteRange >> gtags t
+          _                                          -> gtags t
+
+instance ToTags CodeQL.HigherOrderTerm
```
tclem, line 118: Put it back alphabetically? Just wondering how hard these are going to be to keep track of if/when we add/remove syntax terms.

rewinfrey: I 🔤 these instances in https://github.com/github/semantic/pull/528/commits/439e96feea6f6fc3d826f9944c5e0e4036027b86.

PR #543, opened 2020-04-24
Revamp codegen docs

aymannadeem:
Fixes https://github.com/github/semantic/issues/509


docs/codegen.md
```diff
@@ -104,16 +91,34 @@ instance Foldable Module
...
+
+| Type | JSON | TH-generated code |
+|----------|--------------|------------|
+|Named leaf|<code>{<br>"type": "identifier",<br>"named": true<br>}|<code>data TreeSitter.Python.AST.Identifier a<br>= TreeSitter.Python.AST.Identifier {TreeSitter.Python.AST.ann :: a,<br>TreeSitter.Python.AST.bytes :: text-1.2.3.1:Data.Text.Internal.Text} -- Defined at TreeSitter/Python/AST.hs:10:1<br>instance Show a => Show (TreeSitter.Python.AST.Identifier a) -- Defined at TreeSitter/Python/AST.hs:10:1<br>instance Ord a => Ord (TreeSitter.Python.AST.Identifier a) -- Defined at TreeSitter/Python/AST.hs:10:1<br>instance Eq a => Eq (TreeSitter.Python.AST.Identifier a) -- Defined at TreeSitter/Python/AST.hs:10:1<br>instance Traversable TreeSitter.Python.AST.Identifier -- Defined at TreeSitter/Python/AST.hs:10:1<br>instance Functor TreeSitter.Python.AST.Identifier -- Defined at TreeSitter/Python/AST.hs:10:1<br>instance Foldable TreeSitter.Python.AST.Identifier -- Defined at TreeSitter/Python/AST.hs:10:1<br>instance Unmarshal TreeSitter.Python.AST.Identifier -- Defined at TreeSitter/Python/AST.hs:10:1<br>instance SymbolMatching TreeSitter.Python.AST.Identifier -- Defined at TreeSitter/Python/AST.hs:10:1|
```
tclem, line 98: I don't see closing `</code>` elements, but appears to render OK anyway? Might be nice to have the last TH-generated code be a `<pre>` instead so it flows a bit better.

docs/codegen.md
```diff
@@ -1,45 +1,30 @@
...
+![image](https://user-images.githubusercontent.com/875834/80240750-66d4dd80-8630-11ea-8bbf-4dd3adf65a58.png)
```
tclem, line 23: This diagram is so, so, so helpful. ❤️ ✨ 

What do you think about showing which parts of the process are runtime vs. compile time in the Automated AST Generation column?

docs/codegen.md
```diff
@@ -1,45 +1,30 @@
-@ -1,216 +0,0 @@
 # CodeGen Documentation
 
-CodeGen is the process for auto-generating language-specific, strongly-typed ASTs to be used in [Semantic](https://github.com/github/semantic-code/blob/d9f91a05dc30a61b9ff8c536d75661d417f3c506/design-docs/precise-code-navigation.md).
+CodeGen is the process for auto-generating language-specific, strongly-typed ASTs to be used in [Semantic](https://github.com/github/semantic-code/blob/d9f91a05dc30a61b9ff8c536d75661d417f3c506/design-docs/precise-code-navigation.md). 
+Since it is a critical component of Semantic's language support process, we recommend reading [these docs](https://github.com/github/semantic/blob/master/docs/adding-new-languages.md) first, as they provide an overview of the pipeline CodeGen supports.
```
tclem, line 4: I think we should make all these github links permalinks so that they don't break when we move files around. You can always replace the sha with `master` if you need to, but it's hard to go the other way.

PR #547, opened 2020-05-02
Refactor joinPaths to use pathtype

zhujinxuan:
It is about a function not refactored in  https://github.com/github/semantic/pull/532

### DIFF
--- a/test/Data/Abstract/Path/Spec.hs
+++ b/test/Data/Abstract/Path/Spec.hs
@@ -2,19 +2,22 @@ module Data.Abstract.Path.Spec(spec) where
 
 import Data.Abstract.Path
 import SpecHelpers
+import qualified System.Path as Path
 
 spec :: Spec
 spec = parallel $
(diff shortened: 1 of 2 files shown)


test/Data/Abstract/Path/Spec.hs
```diff
@@ -2,19 +2,22 @@ module Data.Abstract.Path.Spec(spec) where
...
 import Data.Abstract.Path
 import SpecHelpers
+import qualified System.Path as Path
 
 spec :: Spec
 spec = parallel $
   describe "joinPaths" $ do
     it "joins empty paths" $
-      joinUntypedPaths "" "" `shouldBe` "."
+      go Path.currentDir Path.currentDir $ Path.currentDir
```
zhujinxuan, line 11: This is a bit different from the original one.  The original one is
```
Path.currentDir Path.</> Path.relPath "."
```
The current one drops the ` Path.relPath "."`.

PR #552, opened 2020-05-18
Updates unofficial stack configuration information

jkachmar:
I was reading through the README recently and I realized that I hadn't updated the gist that you were pointing to in quite some time.

Since I'm not actively using `semantic` and I'm not likely to fix things unless I'm prodded to do so, I set up a small repository tracking the latest LTS (15.13 at the time of writing) and validate that it works using a simple GitHub Actions workflow.

I'm still a little too lazy to have my repo perform a full build and test run, but this at least should at least provide the user with a build plan that's valid with respect to package versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ cabal v2-test
 cabal v2-run semantic -- --help
 ```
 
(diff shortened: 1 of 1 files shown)


joshvera:
Thanks @jkachmar!

PR #563, opened 2020-06-02
Add 'node_type' field to the Symbol protobuf message

maxbrunsfeld:
See https://github.com/github/semanticd/pull/2638


Dockerfile
```diff
@@ -17,12 +17,12 @@ RUN go get github.com/golang/protobuf/proto && \
...
 
 # Bit of a hack so that proto-lens-protoc actually runs
-COPY --from=haskell /opt/ghc/8.8.3/lib/ghc-8.8.3/* /opt/ghc/8.8.3/lib/ghc-8.8.3/
+COPY --from=haskell /opt/ghc/8.8.1/lib/ghc-8.8.1/* /opt/ghc/8.8.1/lib/ghc-8.8.1/
 
 ENTRYPOINT ["/protobuf/bin/protoc", "-I/protobuf", "--plugin=protoc-gen-haskell=/usr/local/bin/proto-lens-protoc"]
 
 # Build semantic
-FROM haskell:8.8.3 as build
+FROM haskell:8.8.1 as build
```
tclem, line 25: `proto-lens-protoc` seems to only run under 8.8.1. Punting fixing that for another day.

Dockerfile
```diff
@@ -17,12 +17,12 @@ RUN go get github.com/golang/protobuf/proto && \
 COPY --from=haskell /root/.cabal/bin/proto-lens-protoc /usr/local/bin/proto-lens-protoc
 
 # Bit of a hack so that proto-lens-protoc actually runs
-COPY --from=haskell /opt/ghc/8.8.3/lib/ghc-8.8.3/* /opt/ghc/8.8.3/lib/ghc-8.8.3/
+COPY --from=haskell /opt/ghc/8.8.1/lib/ghc-8.8.1/* /opt/ghc/8.8.1/lib/ghc-8.8.1/
```
maxbrunsfeld, line 20: When pairing, @tclem and found that `protoc` did not work with ghc 8.8.3. We didn't investigate the error too deeply though.

PR #565, opened 2020-06-04
Produce syntax_type while tagging, define possible types as enum in proto

tclem:
Defining the types of syntax (e.g. "Function", "Class") semantic cares to tag has been rather adhoc. There are a number of mappings from string to data type representations of this in the code and then further inferences about which types mean _Definition_ vs. _Reference_. This streamlines all that by:

1. Defining a `SyntaxType` enum in our protobuf schema
2. Using the generated datatypes from ☝️ throughout the code
3. Removing a bunch of custom mapping between strings and datatypes

I also made is so that we produce `NodeType` (determines if semantic considers something a definition or a reference) when tagging instead of inferring later. This brings the stack graph and AST based tagging into the same data model.

Finally, there's some cleanup. Tagging always requires the precise parsers (there were some outdated go test still using both) and because there is less mapping between datatypes we can remove some more legacy api types. 🤘


proto/semantic.proto
```diff
@@ -161,12 +162,24 @@ enum NodeType {
   REFERENCE = 4;
 }
 
+enum SyntaxType {
+  UNKNOWN = 0;
```
maxbrunsfeld, line 166: What is `UNKNOWN` needed for?

src/Semantic/Api/Symbols.hs
```diff
@@ -10,137 +10,79 @@
...
-        nodeTypeForKind :: Kind -> NodeType
-        nodeTypeForKind = \case
-          Function       -> DEFINITION
-          Method         -> DEFINITION
-          Class          -> DEFINITION
-          Module         -> DEFINITION
-          Interface      -> DEFINITION
-          Call           -> REFERENCE
-          Type           -> REFERENCE
-          Implementation -> REFERENCE
```
maxbrunsfeld, line 121: ⚡  Great to see this go away.

proto/semantic.proto
```diff
@@ -161,12 +162,24 @@ enum NodeType {
...
+}
+
 message StackGraphNode {
   int64 id = 1;
   string name = 2;
   string line = 3;
-  string kind = 4;
-  Span span = 5;
+  Span span = 4;
+  SyntaxType syntax_type = 5;
```
maxbrunsfeld, line 182: Is it ok to remove `kind` and renumber the fields in the protobuf? Will we need to do some kind of simultaneous rollout of the change on the client and the server?

proto/semantic.proto
tclem: In protobufs, the [default value](https://developers.google.com/protocol-buffers/docs/proto3) for an enum is the first enum value (which must be 0). This makes sure that we can tell the difference between the default and the field being set to something specific.

proto/semantic.proto
tclem: It's only OK b/c we don't use this endpoint at all yet.

proto/semantic.proto
maxbrunsfeld: Oh, that makes sense. I thought that all fields were always 'nullable' in a protobuf, but it looks like that is only true for fields whose values are other messages.

PR #570, opened 2020-06-11
Deprecate diffing graphing

tclem:
Removes support for the `diff` and `graph` commands which currently only work on a la carte syntax terms. This is instead of extracting to new packages (https://github.com/github/semantic/pull/440) and the patch is here when we want to get precise ASTs working with diffs in some future version of semantic.

I feel a little 😢  about removing some of this old and interesting code, but I think the project will be healthier with lugging this functionality around. Looking forward to a fully precise AST world and this sets the stage for completely removing a la carte syntax next.


dcreager:
Haven't done a deep review yet, but I'm :+1: to this.  The great thing about source control is that the code is never truly gone!

PR #585, opened 2020-06-30
fused-effects 1.1

robrix:
This PR updates `semantic` to depend on the new distributive algebras which will ship in `fused-effects` 1.1.

- [x] Depends on #579.
- [x] Depends on a `fused-effects` 1.1.
- [x] Depends on everything else depending on `fused-effects` 1.1.
  - [x] Depends on https://github.com/fused-effects/fused-effects-readline/pull/6
  - [x] Depends on https://github.com/fused-effects/fused-effects-exceptions/pull/15
  - [x] Depends on https://github.com/tree-sitter/haskell-tree-sitter/pull/287
- [x] What is wrong with `haskeline` 0.8? Why are computers so mean??
- [x] Update:
  - [x] `semantic`
  - [x] `semantic-analysis`
  - [x] `semantic-ast`
  - [x] `semantic-codeql`
  - [x] `semantic-core`
  - [x] `semantic-go`
  - [x] `semantic-java`
  - [x] `semantic-json`
  - [x] `semantic-parse`
  - [x] `semantic-php`
  - [x] `semantic-proto`
  - [x] `semantic-python`
  - [x] `semantic-ruby`
  - [x] ~~`semantic-rust`~~ I think this one isn’t set up yet or something
  - [x] `semantic-scope-graph`
  - [x] `semantic-source`
  - [x] `semantic-tags`
  - [x] `semantic-tsx`
  - [x] `semantic-typescript`


robrix:
@patrickt: It looks like `semantic-rust` isn’t set up with a library target. Maybe we could 🍐 on adding one next week?

robrix:
I don’t think I understand the build failures here.

semantic-ast/src/AST/Unmarshal.hs
```diff
@@ -60,10 +59,6 @@ import           TreeSitter.Node as TS
 import           TreeSitter.Parser as TS
 import           TreeSitter.Tree as TS
 
-asks :: Has (Reader r) sig m => (r -> r') -> m r'
-asks f = send (Ask (pure . f))
-{-# INLINE asks #-}
-
```
robrix, line 66: `asks` [is now inlined in `fused-effects` 1.1](http://hackage.haskell.org/package/fused-effects-1.1.0.0/docs/src/Control.Effect.Reader.html#asks), so we don’t have to provide our own definition.

Note that it is _extremely_ likely that this was a red herring in the profiles in the first place; `-fprof-auto` inserts virtual `SCC` annotations around every call to a non-`INLINE`d function, which (among other things) artificially prevents inlining from taking place, and slows down the resulting program. Thus a program which would have cheaply computed a bunch of `ask`s by simply _returning a parameter_ can end up looking extremely expensive in the profile simply because we’ve asked `ghc` to do something a bit silly.

Incidentally, this is contributing significantly to my general distaste for `ghc`’s extant profiling mechanisms.

semantic/semantic.cabal
```diff
@@ -42,6 +42,10 @@ common haskell
     -DBAZEL_BUILD=0
   if (impl(ghc >= 8.8))
     ghc-options: -Wno-missing-deriving-strategies
+  if (impl(ghc >= 8.10))
+    ghc-options:
+      -Wno-missing-safe-haskell-mode
+      -Wno-prepositive-qualified-module
```
robrix, line 48: I’ve also (I think) got everything building with `ghc` 8.10, tho I haven’t e.g. made CI use that yet.

semantic/src/Control/Effect/Interpose.hs
```diff
@@ -1,57 +0,0 @@
-{-# LANGUAGE DeriveFunctor, ExistentialQuantification, FlexibleContexts, FlexibleInstances, GeneralizedNewtypeDeriving, KindSignatures, MultiParamTypeClasses, RankNTypes, StandaloneDeriving, TypeOperators, UndecidableInstances #-}
```
robrix, line 1: 🔥

semantic/src/Control/Effect/Timeout.hs
```diff
@@ -15,7 +14,7 @@ import qualified System.Timeout as System
 --
 -- Any state changes in the action are discarded if the timeout fails.
 timeout :: Has (Lift IO) sig m => Duration -> m a -> m (Maybe a)
-timeout n m = liftWith $ \ ctx hdl
+timeout n m = liftWith $ \ hdl ctx
```
robrix, line 17: These got reordered to match `alg`.

robrix:
The bazel build still needs to be updated to take the new `tree-sitter-*` versions into account. Will merge after that.

robrix:
Alright, builds with both bazel and cabal. Merging 🚀

Issue #574, opened 2020-06-22
Some errors are completely ignored


llelf:
e. g.
```haskell
λ> parseByteString @JSON.Document @Text tree_sitter_json "{\"a\":} [1]"
Right
  (Document
     {ann = "{\"a\":} [1]",
      extraChildren =
        Value
          {getValue =
             R1
               (R1
                  (R1
                     (R1
                        (L1
                           (Object
                              {ann = "{\"a\":} [1]",
                               extraChildren =
                                 [Pair
                                    {ann = "\"a\":} [1]",
                                     value =
                                       Value
                                         {getValue =
                                            L1
                                              (Array
                                                 {ann = "[1]",
                                                  extraChildren =
                                                    [Value
                                                       {getValue =
                                                          R1
                                                            (R1
                                                               (R1
                                                                  (L1
                                                                     (Number
                                                                        {ann =
                                                                           "1",
                                                                         text =
                                                                           "1"}))))}]})},
                                     key =
                                       R1
                                         (String
                                            {ann = "\"a\"",
                                             extraChildren =
                                               Just
                                                 StringContent
                                                 {ann = "a",
                                                  extraChildren =
                                                    []}})}]})))))}})
```
vs
```
$ tree-sitter parse 2.json
(document [0, 0] - [2, 0]
  (object [0, 0] - [0, 10]
    (pair [0, 1] - [0, 10]
      key: (string [0, 1] - [0, 4]
        (string_content [0, 2] - [0, 3]))
      (ERROR [0, 5] - [0, 6])
      value: (array [0, 7] - [0, 10]
        (number [0, 8] - [0, 9])))))
```
version=current master

PR #639, opened 2021-02-02
Proof of concept of "side channel" for diagnostics

alanz:
Replaces `ReaderC` with `StateC` in `Unmarshal`, and checks for diagnostics
at each visited node, kept in the state.

The diagnostic determination is dodgy, but the concept works

See https://gist.github.com/alanz/2ae23a75bba75ed09780d3fa043bd28e

Note: haskell-tree-sitter is a git dep, it needs a
"git submodule update" in the appropriate "dist-newstyle/src" directory.

Related issues: #574 #638 .

Alternative approaches are to extend the query capabilities, as requested in https://github.com/tree-sitter/tree-sitter/issues/606

The `ts_node_has_error` and `ts_node_is_missing_p` calls are not the right ones, the detailed info comes from the logic in 
https://github.com/tree-sitter/tree-sitter/blob/2b0de9dfec62af7e74c319abe80912b02dca74be/lib/src/subtree.c#L846-L894

and

https://github.com/tree-sitter/tree-sitter/blob/2b0de9dfec62af7e74c319abe80912b02dca74be/lib/src/language.c#L67-L80

### DIFF
--- a/semantic-ast/src/AST/Unmarshal.hs
+++ b/semantic-ast/src/AST/Unmarshal.hs
@@ -10,6 +10,7 @@
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE OverloadedStrings #-}
 
 module AST.Unmarshal
 ( parseByteString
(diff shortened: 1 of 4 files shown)


semantic-ast/src/AST/Unmarshal.hs
```diff
@@ -14,6 +14,7 @@
 module AST.Unmarshal
```
erickadalberto, line 15: [\]

erickadalberto (approved):
- [ ] 

\e68a597
\35b96d9
\0b09d4d
\484decd

PR #667, opened 2022-02-25
Move & rename the analyses to group them together

robrix:
This PR just renames the analyses so they're out of the way.

Depends on #666.

rhea

_photo by National Geographic_

### DIFF
--- a/semantic-analysis/semantic-analysis.cabal
+++ b/semantic-analysis/semantic-analysis.cabal
@@ -44,16 +44,17 @@ library
   import: common
   hs-source-dirs: src
   exposed-modules:
+    Analysis.Analysis.Concrete
+    Analysis.Analysis.Exception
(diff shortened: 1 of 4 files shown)


semantic-analysis/semantic-analysis.cabal
```diff
@@ -44,24 +44,24 @@ library
   import: common
   hs-source-dirs: src
   exposed-modules:
+    Analysis.Analysis.Concrete
+    Analysis.Analysis.Exception
+    Analysis.Analysis.Typecheck
```
robrix, line 49: Moving these really just gets them out of the way so they don't look like the mechanisms underlying interpretation. E.g. if we added an interval analysis and also used intervals here and there, we'd have conflicts.

rewinfrey (approved):
nandu

nandu says 👍

PR #668, opened 2022-02-25
Modular abstract interpretation

robrix:
This PR:

- [x] Brings in the modular abstract interpretation mechanisms (e.g. the linking code) that I forgot to bring over following #659.
- [x] Extends the `Exception` domain to support modular abstract interpretation.
- [x] Depends on #666.
- [x] Depends on #667.
- [x] Depends on https://github.com/tclem/lingo-haskell/pull/14 (sort of).

### DIFF
--- a/semantic-analysis/python.tsg
+++ b/semantic-analysis/python.tsg
@@ -10,6 +10,7 @@
 {
   node @this.node
   attr (@this.node) type = "identifier"
+  var @this.text = (source-text @this)
   attr (@this.node) text = (source-text @this)
 }
 
@@ -24,6 +25,7 @@
   edge @this.node -> @id.node
   attr (@id.node) role = "module-name-fragment"
   attr (@this.node -> @id.node) index = (named-child-index @id)
+  attr (@this.node -> @id.node) text = @id.text
 }
 
 (string) @this
--- a/semantic-analysis/python.tsg
+++ b/semantic-analysis/python.tsg
@@ -6,18 +6,23 @@
   attr (@this.node) type = "module"
 }
 
-(import_statement name: (dotted_name (identifier)* @names)) @this
-{
-  node @this.node
-  attr (@this.node) type = "import"
-  attr (@this.node) import = @names
-}
-
 (identifier) @this
 {
   node @this.node
   attr (@this.node) type = "identifier"
   attr (@this.node) text = (source-text @this)
+}
+
+(import_statement) @this
+{
+  node @this.node
+  attr (@this.node) type = "import"
+}
+
+(import_statement name: (dotted_name (identifier) @id)) @this
+{
+  edge @this.node -> @id.node
+  attr (@this.node -> @id.node) index = (named-child-index @id)
 }
 
 (string) @this
--- a/semantic-analysis/python.tsg
+++ b/semantic-analysis/python.tsg
@@ -6,11 +6,11 @@
   attr (@this.node) type = "module"
 }
 
-(import_statement name: (_) @name) @this
+(import_statement name: (dotted_name (identifier)* @names)) @this
 {
   node @this.node
   attr (@this.node) type = "import"
-  attr (@this.node) import = (source-text @name)
+  attr (@this.node) import = @names
 }
 
 (identifier) @this
--- a/semantic-analysis/python.tsg
+++ b/semantic-analysis/python.tsg
@@ -22,6 +22,7 @@
 (import_statement name: (dotted_name (identifier) @id)) @this
 {
   edge @this.node -> @id.node
+  attr (@id.node) role = "module-name-fragment"
   attr (@this.node -> @id.node) index = (named-child-index @id)
 }
 
--- a/semantic-analysis/python.tsg
+++ b/semantic-analysis/python.tsg
@@ -13,9 +13,11 @@
   attr (@this.node) import = (source-text @name)
 }
 
-(identifier) @id
+(identifier) @this
 {
-  node @id.node
+  node @this.node
+  attr (@this.node) type = "identifier"
+  attr (@this.node) text = (source-text @this)
 }
 
 (string) @this
--- a/semantic-analysis/src/Analysis/Analysis/Exception.hs
+++ b/semantic-analysis/src/Analysis/Analysis/Exception.hs
@@ -4,6 +4,7 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
 module Analysis.Analysis.Exception
@@ -17,13 +18,17 @@ module Analysis.Analysis.Exception
 , ExcC(..)
 ) where
 
+import qualified Analysis.Carrier.Statement.State as A
 import qualified Analysis.Carrier.Store.Monovariant as A
 import           Analysis.Effect.Domain
+import           Analysis.Effect.Env (Env)
+import           Analysis.Effect.Store
 import           Analysis.File
 import           Analysis.FlowInsensitive (cacheTerm, convergeTerm)
 import           Analysis.Name
 import           Control.Algebra
 import           Control.Applicative (Alternative (..))
+import           Control.Effect.Labelled
 import           Control.Effect.State
 import qualified Data.Foldable as Foldable
 import           Data.Function (fix)
@@ -58,7 +63,7 @@ exc e = ExcSet mempty (Set.singleton e)
 exceptionTracing
   :: Ord term
   => ( forall sig m
-     .  Has (Dom ExcSet) sig m
+     .  (Has (Env A.MAddr) sig m, HasLabelled Store (Store A.MAddr ExcSet) sig m, Has (Dom ExcSet) sig m, Has A.Statement sig m)
      => (term -> m ExcSet)
      -> (term -> m ExcSet) )
   -> [File term]
@@ -72,14 +77,16 @@ runFile
   :: ( Has (State (A.MStore ExcSet)) sig m
      , Ord term )
   => ( forall sig m
-     .  Has (Dom ExcSet) sig m
+     .  (Has (Env A.MAddr) sig m, HasLabelled Store (Store A.MAddr ExcSet) sig m, Has (Dom ExcSet) sig m, Has A.Statement sig m)
      => (term -> m ExcSet)
      -> (term -> m ExcSet) )
   -> File term
   -> m (File (Set.Set ExcSet))
 runFile eval = traverse run where
   run
-    = convergeTerm (runExcC . fix (cacheTerm . eval))
+    = A.runStatement (const pure)
+    . A.runEnv @ExcSet
+    . convergeTerm (A.runStore @ExcSet . runExcC . fix (cacheTerm . eval))
 
 newtype ExcC m a = ExcC { runExcC :: m a }
   deriving (Alternative, Applicative, Functor, Monad)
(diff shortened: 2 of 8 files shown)


semantic-analysis/.ghci.repl
```diff
@@ -32,9 +32,12 @@
 -- 8.10+
 :seti -Wno-missing-safe-haskell-mode
 :seti -Wno-prepositive-qualified-module
+-- 9.2+
+:seti -Wno-missing-kind-signatures
+:seti -Wno-missing-signatures
```
robrix, line 37: We now support 9.2, but only for `semantic-analysis` (for now).

semantic-analysis/python.tsg
```diff
@@ -6,9 +6,26 @@
...
+  node @this.node
+  attr (@this.node) type = "import"
+}
+
+(import_statement name: (dotted_name (identifier) @id)) @this
+{
+  edge @this.node -> @id.node
+  attr (@id.node) role = "module-name-fragment"
+  attr (@this.node -> @id.node) index = (named-child-index @id)
+  attr (@this.node -> @id.node) text = @id.text
```
robrix, line 28: This lets us import imports 😊

semantic-analysis/src/Analysis/Analysis/Exception.hs
```diff
@@ -47,6 +62,35 @@ exc :: Exception -> ExcSet
...
+
+runFile
+  :: ( Has (State (A.MStore ExcSet)) sig m
+     , Ord term )
+  => ( forall sig m
+     .  (Has (Env A.MAddr) sig m, HasLabelled Store (Store A.MAddr ExcSet) sig m, Has (Dom ExcSet) sig m, Has A.Statement sig m)
+     => (term -> m ExcSet)
+     -> (term -> m ExcSet) )
+  -> File term
+  -> m (File (Module ExcSet))
```
robrix, line 83: Note that this is the first of our analyses to return a `Module` (and thus to support modular analysis). Unfortunately, because we don't yet track exports, we can't do linking properly.

semantic-analysis/src/Analysis/Carrier/Statement/State.hs
```diff
@@ -0,0 +1,43 @@
...
+
+data Message
+  = Import (NonEmpty Text)
+  deriving (Eq, Ord, Show)
+
+
+-- Statement carrier
+
+runStatement :: ([Message] -> a -> m r) -> StatementC m a -> m r
+runStatement k (StatementC m) = runState (k . reverse) [] m
```
robrix, line 35: `Statement` effects just track sequential, imperative directives, currently limited to imports. This collects them into `Message`s which are then returned via this handler.

semantic-analysis/src/Analysis/Carrier/Store/Monovariant.hs
```diff
@@ -83,3 +87,15 @@ instance Has (State (MStore value)) sig m
...
+-- Running
+
+runFiles
+  :: (forall sig m . Has (State (MStore  value)) sig m => File term -> m (File result))
+  -> [File term]
+  -> (MStore value, [File result])
+runFiles runFile
+  = run
+  . runStoreState
+  . traverse runFile
```
robrix, line 101: This convenience wraps up a pattern common among the (0-CFA) analyses into a convenient bundle. It's also a great example of exposing effects while hiding carriers, the essence of abstraction.

semantic-analysis/src/Analysis/Module.hs
```diff
@@ -0,0 +1,27 @@
...
+
+newtype ModuleSet a = ModuleSet { getModuleSet :: Map.Map Name (Module a) }
+
+instance Semigroup (ModuleSet a) where
+  m1 <> m2 = ModuleSet ((link m2 <$> getModuleSet m1) <> (link m1 <$> getModuleSet m2))
+
+link :: ModuleSet a -> Module a -> Module a
+link (ModuleSet ms) m = Module b' (imports m Set.\\ Map.keysSet ms) (exports m) u' where
+  (u', b') = foldl' (\ (u, b) -> resolve u b . exports) (unknown m, body m) (Map.restrictKeys ms (imports m))
+  resolve u b e = (u Set.\\ Map.keysSet e, b . mappend (Map.restrictKeys e u))
```
robrix, line 27: Brought in verbatim from `membrane`. I intend to tidy this up in a future PR, as it's somewhat unreadable as-is.

semantic-analysis/src/Analysis/Syntax.hs
```diff
@@ -22,105 +20,79 @@ module Analysis.Syntax
...
+data Term
+  = Var Name
+  | Noop
+  | Iff Term Term Term
+  | Bool Bool
+  | String Text
+  | Throw Term
+  | Let Name Term Term
+  | Import (NonEmpty Text)
+  deriving (Eq, Ord, Show)
```
robrix, line 62: The tagless final class, and the `Print` and `Interpret` instantiations of it are gone, because defining an `Ord` instance for `Interpret` would have been agonizing. Instead, we parse into a first-order, initial encoding: plain old data.

semantic-analysis/src/Analysis/Syntax.hs
```diff
@@ -134,48 +106,66 @@ let' n v m = do
 
 -- Parsing
 
-parseFile :: Syntax rep => FilePath -> IO (Either (A.JSONPath, String) (Maybe rep))
+parseFile :: (Has (Throw String) sig m, MonadIO m) => FilePath -> m (File Term)
```
robrix, line 109: We also parse into a `File`, which is convenient since we use those for analysis (so that we can service file/line/etc directives).

semantic-analysis/src/Analysis/Syntax.hs
```diff
@@ -134,48 +106,66 @@ let' n v m = do
...
-    ty <- attrs A..: pack "type"
-    node <- parseType attrs ty
+  edges <- o A..: fromString "edges"
+  index <- o A..: fromString "id"
+  o A..: fromString "attrs" >>= A.withObject "attrs" (\ attrs -> do
+    ty <- attrs A..: fromString "type"
+    node <- parseType attrs edges ty
     pure (index, node, node <$ guard (ty == "module")))
+
+parseType :: A.Object -> [A.Value] -> String -> A.Parser (IntMap.IntMap Term -> Term)
```
robrix, line 137: I pulled this and the rest of `parseNode`'s where clause to the top level, mostly to make it tidier and give them type signatures. The only real change otherwise is using `Term` constructors in `parseType` here instead of `Syntax` methods.

semantic-source/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 0.1.0.2
+
+- Support ghc 9.2.
+
```
robrix, line 4: I bumped `semantic-source`, which turned out to be rather confusing, but eh, it's out there now.

robrix (commented):
I intend to fix the export problem in a follow-up PR; this is, therefore, ready for review.

semantic-analysis/src/Analysis/Carrier/Statement/State.hs
```diff
@@ -0,0 +1,43 @@
...
+import qualified Analysis.Effect.Statement as S
+import           Control.Algebra
+import           Control.Carrier.State.Church
+import           Control.Monad.Fail as Fail
+import           Data.List.NonEmpty (NonEmpty)
+import           Data.Text (Text)
+
+-- Messages
+
+data Message
```
patrickt, line 27: newtype here?

semantic-analysis/src/Analysis/Effect/Statement.hs
```diff
@@ -0,0 +1,20 @@
...
+  simport
+, Statement(..)
+) where
+
+import Control.Algebra
+import Data.Kind as K
+import Data.List.NonEmpty (NonEmpty)
+import Data.Text
+
+-- Statement effect
```
patrickt, line 14: Can you expand a little on this? Where is this intended to be used, etc?

semantic-analysis/src/Analysis/Module.hs
patrickt: Yeah, it’s a bit dense. Would you mind taking a run at it now? Even pulling out and naming the fold function would help a bit.

semantic-analysis/src/Analysis/Syntax.hs
patrickt: Plain old data is (usually) my favorite kind of data!

semantic-analysis/src/Analysis/Syntax.hs
```diff
@@ -22,105 +20,79 @@ module Analysis.Syntax
...
+evalModule0 :: (Has (Env addr) sig m, HasLabelled Store (Store addr val) sig m, Has (Dom val) sig m) => Term -> m (Module val)
+evalModule0 i = S.runStatement mk (eval0 i) where
+  mk msgs b = pure (Module (const b) (Set.fromList (map (\ (S.Import cs) -> name (Text.intercalate (pack ".") (toList cs))) msgs)) mempty mempty)
 
-  let_ n v b = Interpret (\ eval -> do
-    v' <- eval v
-    let' n v' (eval (b (Interpret (pure (pure v'))))))
+evalModule :: (Has (Env addr) sig m, HasLabelled Store (Store addr val) sig m, Has (Dom val) sig m) => (Term -> S.StatementC m val) -> (Term -> m (Module val))
+evalModule f i = S.runStatement mk (eval f i) where
+  mk msgs b = pure (Module (const b) (Set.fromList (map (\ (S.Import cs) -> name (Text.intercalate (pack ".") (toList cs))) msgs)) mempty mempty)
```
patrickt, line 95: Some let-bindings here would go a long way.

semantic-analysis/src/Analysis/Syntax.hs
```diff
@@ -134,48 +106,66 @@ let' n v m = do
...
-  contents <- B.readFile path
-  pure $ snd <$> A.eitherDecodeWith A.json' (A.iparse parseGraph) contents
-
-parseGraph :: Syntax rep => A.Value -> A.Parser (IntMap.IntMap rep, Maybe rep)
+  contents <- liftIO (B.readFile path)
+  case (A.eitherDecodeWith A.json' (A.iparse parseGraph) contents) of
+    Left  (_, err)       -> throwError err
+    Right (_, Nothing)   -> throwError "no root node found"
+    -- FIXME: this should get the path to the source file, not the path to the JSON.
+    -- FIXME: this should use the span of the source file, not an empty span.
```
patrickt, line 116: Do you want to fix these now or later?

semantic-analysis/src/Analysis/Carrier/Statement/State.hs
robrix: Expectation is that there'll be other message types, but in  the meantime, yeah, good call.

semantic-analysis/src/Analysis/Effect/Statement.hs
robrix: 👍🏻

semantic-analysis/src/Analysis/Syntax.hs
robrix: Later, definitely.

semantic-analysis/src/Analysis/Syntax.hs
robrix: I think I'm going to 🔥 them instead, as there's no way to provide the body, unknowns, and exports independent of the domain type anyway.

semantic-analysis/src/Analysis/Syntax.hs
robrix: It's just, it was so nice! But too clever by half.

patrickt (approved):

robrix:
not today I’ve got LEGS motherfucker

PR #679, opened 2022-04-20
Disable the bazel CI action for the time being.

robrix:
This PR does what it says on the tin.


hendrikvanantwerpen (approved):
That is one short pr 😂. Also that branch name... I'd never imagine that is allowed 🤷‍♂️.

Anyway, LGTM

robrix:
Me neither FWIW! But I appear to have got away with it…

PR #687, opened 2022-07-14
Drop pathtype, part one

robrix:
This PR drops `semantic-source`'s dependency on `pathtype`, selecting languages from `FilePath`s instead. I'll release the new version once this PR is approved.

Our other packages get `semantic-source` from `cabal`, and are constrained to exclude 0.2+, so they shouldn't be disturbed by this change. Updating them to drop the `pathtype` dependency is postponed to a later PR.

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -66,7 +66,7 @@ jobs:
         cabal v2-run   --project-file=cabal.project.ci semantic-tags:test
         cabal v2-run   --project-file=cabal.project.ci semantic-tsx:test
         cabal v2-run   --project-file=cabal.project.ci semantic-typescript:test
-        cabal v2-run   --project-file=cabal.project.ci semantic-source:test
+        cd semantic-source && cabal v2-run   --project-file=cabal.project.ci semantic-source:test
 
     - name: Write out cache
       run: ./cabal-cache sync-to-archive --threads=2 --archive-uri=dist-cache
--- a/cabal.project
+++ b/cabal.project
@@ -15,7 +15,6 @@ packages: semantic
           semantic-ruby
           semantic-rust
           semantic-scope-graph
-          semantic-source
           semantic-tags
           semantic-tsx
           semantic-typescript
--- a/semantic-source/src/Source/Language.hs
+++ b/semantic-source/src/Source/Language.hs
@@ -109,34 +109,34 @@ forPath path =
 
 languageToText :: Language -> T.Text
 languageToText = \case
-  Unknown -> "Unknown"
-  CodeQL -> "CodeQL"
-  Go -> "Go"
-  Haskell -> "Haskell"
-  Java -> "Java"
+  Unknown    -> "Unknown"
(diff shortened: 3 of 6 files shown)


.github/workflows/haskell.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         cabal v2-run   --project-file=cabal.project.ci semantic-tags:test
         cabal v2-run   --project-file=cabal.project.ci semantic-tsx:test
         cabal v2-run   --project-file=cabal.project.ci semantic-typescript:test
-        cabal v2-run   --project-file=cabal.project.ci semantic-source:test
+        cd semantic-source && cabal v2-run   --project-file=cabal.project.ci semantic-source:test
```
robrix, line 69: We're excluding this from the top-level project now, so we `cd` in first.

cabal.project
```diff
@@ -15,7 +15,6 @@ packages: semantic
           semantic-ruby
           semantic-rust
           semantic-scope-graph
-          semantic-source
```
robrix, line 18: Here, and in `cabal.project.ci`, we remove `semantic-source` from the top-level project, ensuring that it'll be fetched from hackage. This makes it possible to factor out PRs like this one, at the cost of making it less convenient to change things across both `semantic` (or any other local package) and `semantic-source`.

For this PR I think that's worth it; for the next one I might end up reinstating this. But I wanted to factor the PRs out to begin with to limit scope and to be sure we're handling versioning appropriately.

semantic-source/src/Source/Language.hs
```diff
@@ -96,47 +94,47 @@ knownLanguage = (/= Unknown)
...
 
-forPath :: Path.PartClass.AbsRel ar => Path.File ar -> Language
+forPath :: FilePath -> Language
 forPath path =
   let spurious lang = lang `elem` [ "Hack" -- .php files
                                   , "GCC Machine Description" -- .md files
                                   , "XML" -- .tsx files
                                   ]
-      allResults = Lingo.languageName <$> Lingo.languagesForPath (Path.toString path)
+      allResults = Lingo.languageName <$> Lingo.languagesForPath path
```
robrix, line 103: This is the only real code change.

robrix (commented):
Ready for review.

cabal.project
dcreager: > at the cost of making it less convenient to change things across both

To clarify, this means that the process would be to merge the PR that touches the "downstream" package, cut a new release and deploy it to hackage, and _then_ open the PR in the "upstream" package?

(Not meaning to suggest a different approach, just making sure I understand how the process would be impacted.)

semantic-source/src/Source/Language.hs
dcreager: :smiling_face_with_tear:

dcreager (approved):

robrix:
[Released](https://hackage.haskell.org/package/semantic-source-0.2.0.0).

PR #689, opened 2022-07-15
Drop pathtype, part two

robrix:
This PR removes uses of and dependencies on `pathtype` from the rest of the repo.

Depends on #687. In particular, CI won't succeed until we publish `semantic-source` 0.2 to hackage.

### DIFF
--- a/semantic-analysis/src/Analysis/Blob.hs
+++ b/semantic-analysis/src/Analysis/Blob.hs
@@ -8,13 +8,11 @@ module Analysis.Blob
   , nullBlob
   ) where
 
-import           Analysis.File as A
-import           Analysis.Reference as A
-import           Data.Aeson
-import           Source.Language as Language
-import           Source.Source as Source
-import qualified System.Path as Path
-import qualified System.Path.PartClass as Path.PartClass
+import Analysis.File as A
+import Analysis.Reference as A
+import Data.Aeson
+import Source.Language as Language
+import Source.Source as Source
 
 -- | The source, path information, and language of a file read from disk.
 data Blob = Blob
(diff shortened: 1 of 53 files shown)


semantic-analysis/src/Analysis/Blob.hs
```diff
@@ -8,13 +8,11 @@ module Analysis.Blob
...
-import           Data.Aeson
-import           Source.Language as Language
-import           Source.Source as Source
-import qualified System.Path as Path
-import qualified System.Path.PartClass as Path.PartClass
+import Analysis.File as A
+import Analysis.Reference as A
+import Data.Aeson
+import Source.Language as Language
+import Source.Source as Source
```
robrix, line 15: Expect lots of this sort of thing due to autoformatter. Apologies.

semantic-analysis/src/Analysis/Blob.hs
```diff
@@ -25,27 +23,27 @@ data Blob = Blob
 instance FromJSON Blob where
   parseJSON = withObject "Blob" $ \b -> do
     src <- b .: "content"
-    Right pth <- fmap Path.parse (b .: "path")
+    pth <- b .: "path"
```
robrix, line 26: Simpler terms.

semantic-analysis/src/Analysis/Blob.hs
```diff
@@ -25,27 +23,27 @@ data Blob = Blob
...
     lang <- b .: "language"
     let lang' = if knownLanguage lang then lang else Language.forPath pth
-    pure (fromSource (pth :: Path.AbsRelFile) lang' src)
+    pure (fromSource pth lang' src)
 
 
 -- | Create a Blob from a provided path, language, and UTF-8 source.
 -- The resulting Blob's span is taken from the 'totalSpan' of the source.
-fromSource :: Path.PartClass.AbsRel ar => Path.File ar -> Language -> Source -> Blob
+fromSource :: FilePath -> Language -> Source -> Blob
```
robrix, line 34: Simpler types.

semantic-analysis/src/Analysis/Blob.hs
```diff
@@ -25,27 +23,27 @@ data Blob = Blob
...
 blobLanguage = A.fileBody . blobFile
 
-blobPath :: Blob -> Path.AbsRelFile
+blobPath :: Blob -> FilePath
 blobPath = A.refPath . A.fileRef . blobFile
 
 -- | Show FilePath for error or json outputs.
 blobFilePath :: Blob -> String
-blobFilePath = Path.toString . blobPath
+blobFilePath = blobPath
```
robrix, line 46: I opted not to eliminate this in favour of `blobPath` because I didn't want to get dragged into that refactoring in the same PR. Still, it would be nice.

semantic-codeql/test/PreciseTest.hs
```diff
@@ -1,25 +1,28 @@
...
-import           Test.Tasty
 import qualified System.Path.Fixture as Fixture
+import           Test.Tasty
 
 main :: IO ()
 main = do
 #if BAZEL_BUILD
   rf <- Fixture.create
-  let ?project = Path.relDir "external/tree-sitter-ql"
+  let ?project = "external/tree-sitter-ql"
```
robrix, line 20: We aren't currently building with bazel, but I did my best to keep these changes bazel-friendly at least.

semantic/test/Semantic/CLI/Spec.hs
```diff
@@ -34,31 +33,31 @@ testTree = testGroup "Semantic.CLI"
...
 renderDiff ref new = unsafePerformIO $ do
   let check p = do
-        exists <- Path.doesFileExist (Path.absRel p)
+        exists <- doesFileExist p
         unless exists (throwIO (userError ("Can't find path " <> p)))
 
   check ref
   check new
-  useJD <- (Path.hasExtension ".json" (Path.absRel ref) &&) <$> fmap isJust (Path.findExecutable "jd")
+  useJD <- (takeExtension ref == ".json" &&) <$> fmap isJust (findExecutable "jd")
```
robrix, line 41: `pathtype` generally mimics `filepath`'s API pretty closely, so most of the changes either just drop conversions to/from strings, or drop the module prefix and use the mimicked symbol. In this case, `pathtype` does something completely different for some reason, so there's some actual logic to this change.

I did double-check that `takeExtension` returns the `.`.

robrix (commented):
Ready for review!

patrickt (approved):

PR #690, opened 2022-07-21
Demo

robrix:
Some last tweaks to polish off the exception tracing demo.

### DIFF
--- a/semantic-analysis/src/Analysis/Syntax.hs
+++ b/semantic-analysis/src/Analysis/Syntax.hs
@@ -247,6 +247,6 @@ analyzeFile path analyze = do
   analyze eval src file
 
 parseToTerm :: (Algebra sig m, MonadIO m) => FilePath -> m (Source.Source, File Term)
-parseToTerm path = do
-  parsed <- runThrow @String (parseFile path)
+parseToTerm jsonPath = do
+  parsed <- runThrow @String (parseFile jsonPath)
   either (liftIO . throwIO . ErrorCall) pure parsed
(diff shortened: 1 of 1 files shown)


semantic-analysis/src/Analysis/Syntax.hs
```diff
@@ -129,12 +129,12 @@ letrec n m = do
...
-  contents <- liftIO (B.readFile path)
-  let sourcePath = replaceExtensions path "py"
-  -- FIXME: this is comprised of several terrible assumptions.
-  sourceContents <- Source.fromUTF8 . B.toStrict <$> liftIO (B.readFile sourcePath)
+parseFile :: (Has (Throw String) sig m, MonadIO m) => FilePath -> FilePath -> m (Source.Source, File Term)
+parseFile srcPath jsonPath = do
+  contents <- liftIO (B.readFile jsonPath)
+  -- FIXME: get this from the JSON itself (cf https://github.com/tree-sitter/tree-sitter-graph/issues/69)
+  let sourcePath = replaceExtensions jsonPath "py"
+  sourceContents <- Source.fromUTF8 . B.toStrict <$> liftIO (B.readFile srcPath)
```
robrix, line 137: This is very slightly less gross than it had been, but still pretty gross. Basically what it amounts to is a way to allow the JSON path and the loaded source file to vary: in the demo's case, I wanted both `b.py` and `b2.py` to load their respective sources correctly while still obtaining the module name `b`.

As of this PR, we're making one fewer assumption than we had been, so I'm content with it for the moment, but a real solution will be needed at some point.

PR #714, opened 2024-03-11
Dead code analysis

robrix:
Run it like so:

```haskell
deadCodeFlowInsensitive Analysis.Syntax.eval Analysis.Syntax.subterms [Analysis.File.File (Analysis.Reference.Reference "" (Source.Span.Span (Source.Span.Pos 0 0) (Source.Span.Pos 1 1))) (Analysis.Syntax.String (Data.String.fromString ""))]
```

to get a result like so:

```
( fromList []
, …
, …
)
```

which shows us (by its exclusion) that the entire input program is live.

### DIFF
--- a/semantic-analysis/cabal.project
+++ b/semantic-analysis/cabal.project
@@ -2,3 +2,4 @@
 
 -- Local packages
 packages: .
+          ../semantic-source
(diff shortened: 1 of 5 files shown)


robrix:
NB: we currently evaluate function bodies at their definition sites, which means that all code ends up appearing live. I'm switching focus to another part of the system for the moment, so I'm postponing that issue to another PR.

BekaValentine (approved):

PR #715, opened 2024-03-15
(Quasi)ABTs for syntax types

robrix:
Props to @BekaValentine for a discussion of the advantages of the approach for typed and untyped syntax. As yet we're only employing untyped syntax, but it's looking pretty reasonable.

### DIFF
--- a/semantic-analysis/src/Analysis/Syntax.hs
+++ b/semantic-analysis/src/Analysis/Syntax.hs
@@ -1,270 +1,83 @@
-{-# LANGUAGE ExistentialQuantification #-}
-{-# LANGUAGE FlexibleContexts #-}
-{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE LambdaCase #-}
-{-# LANGUAGE MultiParamTypeClasses #-}
-{-# LANGUAGE RankNTypes #-}
-{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE QuantifiedConstraints #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
 module Analysis.Syntax
-( -- * Terms
+( -- * Syntax
   Term(..)
-, subterms
-  -- * Abstract interpretation
-, eval0
-, eval
-  -- * Macro-expressible syntax
-, let'
-, letrec
-  -- * Parsing
-, parseFile
-, parseGraph
-, parseNode
-  -- * Debugging
-, analyzeFile
-, parseToTerm
+  -- * Vectors
+, Nat(..)
+, N0
+, N1
+, N2
+, N3
+, Vec(..)
+, toList
 ) where
 
-import qualified Analysis.Carrier.Statement.State as S
-import           Analysis.Effect.Domain
-import           Analysis.Effect.Env (Env, bind, lookupEnv)
-import           Analysis.Effect.Store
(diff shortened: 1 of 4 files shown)


semantic-analysis/src/Analysis/Syntax/Python.hs
```diff
@@ -0,0 +1,198 @@
...
+    (Function' n1 as1, Function' n2 as2) -> compare n1 n2 <> compare as1 as2
+    (Function' _ _, _)                   -> LT
+    (Call', Call')                       -> EQ
+    (Call', _)                           -> LT
+    (ANil', ANil')                       -> EQ
+    (ANil', _)                           -> LT
+    (ACons', ACons')                     -> EQ
+    (ACons', _)                          -> LT
+    (Locate' s1, Locate' s2)             -> compare s1 s2
+    (Locate' _, _)                       -> LT
```
robrix, line 162: We can't derive these because `GHC.Generics` doesn't play well with `GADTs` (or the other way around, I suppose). This is substantially worse than what we'd get using standard `Fix`-based terms or even plain old data.

semantic-analysis/src/Analysis/Syntax.hs
```diff
@@ -1,270 +1,53 @@
...
-  | Call Term [Term]
-  | Locate Span Term
-  deriving (Eq, Ord, Show)
+-- Syntax
 
-infixl 1 :>>
+-- | (Currently) untyped term representations.
+data Term sig v
+  = Var v
+  | Term (sig (Term sig v))
```
robrix, line 20: A free monad by any other name

semantic-analysis/src/Analysis/Syntax.hs
```diff
@@ -1,270 +1,53 @@
...
-  Let _ v b      -> subterms v <> subterms b
-  a :>> b        -> subterms a <> subterms b
-  Import _       -> mempty
-  Function _ _ b -> subterms b
-  Call f as      -> subterms f <> foldMap subterms as
-  Locate _ b     -> subterms b
+instance (Eq (sig (Term sig v)), Eq v) => Eq (Term sig v) where
+  Var v1  == Var v2  = v1 == v2
+  Term s1 == Term s2 = s1 == s2
+  _       == _       = False
```
robrix, line 25: The instance head can be cleaned up (and avoid the use of `UndecidableInstances`) using `QuantifiedConstraints`. However:

- for `Ord` you end up needing to supply quantified constraints for both `Eq` and `Ord`, and
- it was causing spurious CI failures for ghc9.2 and I really can't be bothered

semantic-analysis/src/Analysis/Syntax.hs
```diff
@@ -1,270 +1,53 @@
...
-parseGraph :: A.Value -> A.Parser (Graph, Maybe Term)
-parseGraph = A.withArray "nodes" $ \ nodes -> do
-  (untied, First root) <- fold <$> traverse parseNode (V.toList nodes)
-  -- @untied@ is an intmap, where the keys are graph node IDs and the values are functions from the final graph to the representations of said graph nodes. Likewise, @root@ is a function of the same variety, wrapped in a @Maybe@.
-  --
-  -- We define @tied@ as the fixpoint of the former to yield the former as a graph of type @Graph@, and apply the latter to said graph to yield the entry point, if any, from which to evaluate.
-  let tied = fix (\ tied -> ($ Graph tied) <$> untied)
-  pure (Graph tied, ($ Graph tied) <$> root)
+foldTerm :: Functor sig => (v -> r) -> (sig r -> r) -> (Term sig v -> r)
+foldTerm var sig = mendlerTerm var (\ k -> sig . fmap k)
```
robrix, line 38: Here be recursion schemes 🐉

robrix (commented):
Ready for review.

BekaValentine (approved):

PR #716, opened 2024-03-20
Ingest Python

robrix:


semantic-analysis/src/Analysis/Syntax/Python.hs
```diff
@@ -1,82 +1,122 @@
...
+import           System.FilePath (takeBaseName)
 
 -- Syntax
 
-type Term = T.Term Python Name
+data Term
+  = Module (Py.Module Py.SrcSpan)
+  | Statement (Py.Statement Py.SrcSpan)
+  | Expr (Py.Expr Py.SrcSpan)
+  deriving (Eq, Ord, Show)
```
robrix, line 40: We wrap `language-python` AST nodes up in constructors of a single datatype so as to avoid having to define our _own_ AST for Python and a complicated and error-prone function for copying things from the `language-python` AST in. This type isn't recursive at all; we just wrap the current layer, and then wrap the next layer on the fly in recursive functions like `subterms` and `eval`.

Somewhat to my surprise, the `language-python` AST (and, I suppose, Python's own syntax) is regular enough to avoid making this very much more complex. There are lots of other types representing identifiers, arguments, parameters, and all the other, weirder bits of syntax, but most of them are auxiliary—not things we would evaluate.

semantic-analysis/src/Analysis/Syntax/Python.hs
```diff
@@ -1,82 +1,122 @@
...
-  | Let Name t t
-  | t :>> t
-  | Import (NonEmpty Text)
-  | Function Name [Name] t
-  | Call t [t]
-  | Locate Span t
-  deriving (Eq, Foldable, Functor, Ord, Show, Traversable)
-
-infixl 1 :>>
+-- | Non-generic production of the recursive set of subterms.
```
robrix, line 42: 😭 we can't get this generically for all languages any more using this approach.

semantic-analysis/src/Analysis/Syntax/Python.hs
```diff
@@ -1,82 +1,122 @@
...
+  -- FIXME: Py.RaiseV2
+  -- FIXME: whatever the second field is
+  Statement (Py.StmtExpr e _)           -> subterms (Expr  e)
+  Statement (Py.Fun _ _ _ ss _)         -> suite ss
+  -- FIXME: include 'subterms' of any default values
+  Expr (Py.Call f as _)                 -> subterms (Expr f) <> foldMap (\case { Py.ArgExpr e _ -> subterms (Expr e) ; _ -> Set.empty }) as
+  -- FIXME: support keyword args &c.
+  _                                     -> Set.empty -- TBD, and terminals
+  where
+  suite = foldMap (subterms . Statement)
```
robrix, line 59: Note that various terminal nodes (literals, etc) are all (correctly) handled by the fallback case—but so are nonterminals we just haven't got around to yet.

I couldn't think of any generic way of implementing this which wouldn't end up being more trouble than hand-writing (and, regrettably, -maintaining) this. Oh, except I just did: an abstract interpretation of a program ends up being a kind of complicated fold. So if we were to define a domain modelling a set of terms, we could use one abstract interpretation to compute the set(s) of terms to use for dead code analysis in a second pass 😝

semantic-analysis/src/Analysis/Syntax/Python.hs
```diff
@@ -1,82 +1,122 @@
...
+  Statement (Py.Fun n ps _r ss sp) -> let ps' = mapMaybe (\case { Py.Param n _ _ _ -> Just (ident n) ; _ -> Nothing }) ps in setSpan sp $ letrec (ident n) (dabs ps' (foldr (\ (p, a) m -> let' p a m) (suite ss) . zip ps'))
+  Expr (Py.Var n sp) -> setSpan sp $ let n' = ident n in lookupEnv n' >>= maybe (dvar n') fetch
+  Expr (Py.Bool b sp) -> setSpan sp $ dbool b
+  Expr (Py.Strings ss sp) -> setSpan sp $ dstring (pack (mconcat ss))
+  Expr (Py.Call f as sp) -> setSpan sp $ do
+    f' <- eval (Expr f)
+    as' <- traverse eval (mapMaybe (\case { Py.ArgExpr e _ -> Just (Expr e) ; _ -> Nothing }) as)
+    -- FIXME: support keyword args &c.
+    dapp f' as'
+  _ -> fail "TBD"
```
robrix, line 97: We handle one fewer case: `Let` is gone. However, `Let` was never produced from actual Python code using the TSG generator, so it's unclear why it was in there to begin with.

We also handle statement expressions which I think may not have been handled? And the failure case is explicit here, instead of being hidden in the term ingestion machinery like it had been.

semantic-analysis/src/Analysis/Syntax/Python.hs
```diff
@@ -1,82 +1,122 @@
...
   where
-  setSpan s r = r{ refSpan = s }
+  setSpan s = case fromSpan s of
+    Just s -> local (\ r -> r{ refSpan = s })
+    _      -> id
```
robrix, line 101: `Locate` cases are no longer so bespoke, and are instead sprinkled around throughout the evaluation. I mean, they probably were before anyway, but this is a little more explicit about it.

We could probably call this once, outside the `\case`, using `language-python`'s typeclass for annotated nodes, but I couldn't be bothered.

semantic-ast/semantic-ast.cabal
```diff
@@ -76,7 +76,7 @@ library
     , semantic-source      ^>= 0.2
     , tasty                ^>= 1.2.3
     , tasty-hunit          ^>= 0.10.0.2
-    , template-haskell      >= 2.15 && < 2.19
+    , template-haskell      >= 2.15 && < 2.22
```
robrix, line 79: Don't mind me, just restoring compilation under newer compilers which CI doesn't exercise yet.

robrix (commented):
Ready for review.

BekaValentine (approved):

